PR #2, opened 2023-02-16
Visualize Entities in a diagram

martin-fleck-at:
- Add diagram language to Langium grammar to store notation data
-- Do not export any elements from diagram (scoping)
-- Extend serializer to handle diagram models

- Provide GLSP server for diagram documents on port 5007
-- Add server to language server process to share LSP services
-- Use LSP services to load diagram document (CrossModelStorage)
-- Transform diagram and semantic models to GModel (GModelFactory)
-- Configure diagram with default element type IDs

- Add GLSP client to Theia for diagram documents (*.diagram.cm)
-- Listen to server messages to startup contribution
-- Use default model elements for element types for rendering
-- Provide style for entity nodes (different background)

- Add diagrams to workspace example

- Minor fix to trigger validation on initial build (WorkspaceManager)

https://dev.azure.com/x-breeze/CrossModel/_workitems/edit/704


planger (approved):
Thanks, looks good to me! :+1:

PR #3, opened 2023-02-17
Add form editor and centralized model service

planger:
* Add model service that allows to open, close, upate, and save models
* Add JSON RPC editor that exposes the model service API to remote clients
* Add Theia frontend/backend-service that connects to the model server
* Add form editor based on that service that shows relationships and entities
* Enable editing and saving in the form editor

Please note that the code in `extensions/crossmodel-lang/src/model-server` is meant to be entirely generic (i.e. independent of crossmodel) to potentially allow for open-sourcing this generic functionality and benefit from potential enhancements and hardening in other use cases, if this is desired.

### DIFF
--- a/applications/electron-app/package.json
+++ b/applications/electron-app/package.json
@@ -31,6 +31,7 @@
     "@crossbreeze/core": "0.0.0",
     "@crossbreeze/glsp-client": "0.0.0",
     "@crossbreeze/product": "0.0.0",
+    "@crossbreeze/form-client": "0.0.0",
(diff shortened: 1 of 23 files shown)


martin-fleck-at (approved):
Change looks good to me and allows the user to adapt the source and target of a relationship and the description of an entity through a very simple form, thus covering all aspects of https://dev.azure.com/x-breeze/CrossModel/_workitems/edit/711

Great work, Philip! I especially like the generalization, I'm sure that will come in handy in many places!

PR #4, opened 2023-03-02
Add ChangeBounds, CreateEdge, and Delete operation to GLSP server

martin-fleck-at:
- Add operation handlers for change bounds, create edge and delete
- Add command implementation that uses text as state management
- Add support for negative coordinates

- Move serialization behavior from ModelService to ModelServer
- Make ModelService synchronous until we have async command API in GLSP

Minor:
- Add browser build for application to test integration
- Upgrade to latest Theia community release
- Add property to specify logging behavior of language server

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ If you also made changes to the backend, you can close and restart the appliacti
 
 To package the application use
 
-    yarn electron package
+    yarn theia:electron package
 
 Depending on the platform, this will produce an executable or an installer for the application under `applications/electron-app/dist`.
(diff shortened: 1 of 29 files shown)


planger (commented):
Awesome, thank you very much Martin!
Creating references and moving elements in the diagram seems to work great. However, changing the size of entities didn't work for me. Is that the same for you?

martin-fleck-at:
@planger Thank you for the review! You are right, I missed that case and fixed it in a follow-up commit. With that everything should work as expected.

planger (approved):
Thanks, looks good to me! :+1:

PR #9, opened 2023-04-18
Feature/attributes graphical view

ozcanxbreeze:
Started working on the userstory USER STORY 705.

Done:
- Built a gmodel class named [GEntityNode.ts](https://github.com/CrossBreezeNL/crossmodel/compare/main...feature/attributes-graphical-view#diff-a14bfbe748820a1051bd51c88f9cd96b049fbf0af7097bc512891723acd6eb2a) to build sever side entity nodes.
- Built RectangularNodeView in [views.tsx](https://github.com/CrossBreezeNL/crossmodel/compare/main...feature/attributes-graphical-view#diff-138e8db9222d799a38945d862fd2c2dc44960f7962d2ed8f42be6a1bac8657c5) to render an entity node server side.
- Added css

Still missing:
- Documentation (WIP)
- Test (How are we going to do this?)

### DIFF
--- a/extensions/crossmodel-lang/src/glsp-server/model/builders/GEntityNode.ts
+++ b/extensions/crossmodel-lang/src/glsp-server/model/builders/GEntityNode.ts
@@ -5,71 +5,71 @@ import { ArgsUtil, GCompartment, GLabel, GNode, GNodeBuilder } from '@eclipse-gl
 import { DiagramNode } from '../../../language-server/generated/ast';
 
 export class GEntityNode extends GNode {
-   override type = 'node:entity';
+    override type = 'node:entity';
 
-   static override builder(): GEntityNodeBuilder {
-      return new GEntityNodeBuilder(GEntityNode).type('node:entity');
-   }
+    static override builder(): GEntityNodeBuilder {
+        return new GEntityNodeBuilder(GEntityNode).type('node:entity');
+    }
 }
 
 export class GEntityNodeBuilder extends GNodeBuilder {
-   addNode(node: DiagramNode): this {
-      // Get the reference that the DiagramNode holds to the Entity in the .langium file.
-      const entityRef = node.semanticElement.ref;
+    addNode(node: DiagramNode): this {
+        // Get the reference that the DiagramNode holds to the Entity in the .langium file.
+        const entityRef = node.semanticElement.ref;
 
-      // Options which are the same for every node
-      this.addCssClasses('diagram-node', 'entity').layout('vbox').addArgs(ArgsUtil.cornerRadius(3));
+        // Options which are the same for every node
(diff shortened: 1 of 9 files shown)


extensions/crossmodel-lang/src/glsp-server/model/builders/GEntityNode.ts
```diff
@@ -0,0 +1,75 @@
...
+      // We need the id before we can build the label and childeren.
+      if (this.id === undefined) {
+         throw new Error('Add id to builder before adding the node reference.');
+      }
+
+      // Add the label/name of the node
+      const label = GCompartment.builder()
+         .layout('hbox')
+         .addLayoutOption('hAlign', 'center')
+         .addCssClass('header-container-diagram-node')
```
harmen-xb, line 32: I think heder-compartment would be a better css class name (just like with attributes.

packages/glsp-client/src/browser/crossmodel-diagram-module.ts
```diff
@@ -3,19 +3,24 @@
...
+import { EntityNodeView } from '../views';
+import { EntityNode } from '../model';
 
 const crossModelDiagramModule = new ContainerModule((bind, unbind, isBound, rebind) => {
    rebind(TYPES.ILogger).to(ConsoleLogger).inSingletonScope();
    rebind(TYPES.LogLevel).toConstantValue(LogLevel.warn);
    const context = { bind, unbind, isBound, rebind };
    configureDefaultModelElements(context);
+
+   configureModelElement(context, 'node:entity', EntityNode, EntityNodeView);
```
harmen-xb, line 23: Can you add some comment here what this does?

ozcanxbreeze: added in the code

extensions/crossmodel-lang/src/glsp-server/model/builders/GEntityNode.ts
ozcanxbreeze: done

packages/glsp-client/src/model.ts
```diff
@@ -0,0 +1,7 @@
+/********************************************************************************
+ * Copyright (c) 2023 CrossBreeze.
+ ********************************************************************************/
+
+import { RectangularNode } from 'sprotty/lib';
+
+export class EntityNode extends RectangularNode {}
```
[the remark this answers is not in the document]
ozcanxbreeze: You can put the property values of the client side model here. The way it is implemented, it is useless. I put it here as an example on how to do it, and we wont have to look up how to set up the file when we really do need it. I can delete it if it is not unwanted.

harmen-xb (approved):
Looks good!

PR #21, opened 2023-06-26
Allow adding existing entities to a diagram

martin-fleck-at:
- Create package '@crossbreeze/protocol' for shared interfaces -- Move serializable model definitions into shared package -- Move ModelServer API into shared package

- Add possibility to DnD entities from Explorer to diagram -- Create custom operation in shared GLSP protocol

- Add possibility to add entity through command palette (Ctrl+Space)

add-entities-to-diagram


extensions/crossmodel-lang/src/glsp-server/command-palette/add-entity-action-provider.ts
ozcanxbreeze, line 1: Functionally looks good. Maybe a line of comment to describe in general what is happening in the file or what the purpose is.

extensions/crossmodel-lang/src/glsp-server/handler/add-entity-operation-handler.ts
ozcanxbreeze, line 1: Functionally looks good. Maybe a line of comment to describe in general what is happening in the file or what the purpose is.

small bug, multiple instances of the same node get called 12345 instead of changing the last number, findAvailableNodeName

extensions/crossmodel-lang/src/glsp-server/handler/drop-entity-operation-handler.ts
ozcanxbreeze, line 1: Functionally looks good. Maybe a line of comment to describe in general what is happening in the file or what the purpose is.

small bug, multiple instances of the same node get called 12345 instead of changing the last number, findAvailableNodeName

packages/glsp-client/src/browser/crossmodel-diagram-manager.ts
ozcanxbreeze, line 1: Functionally looks good. Maybe a line of comment to describe in general what is happening in the file or what the purpose is.

packages/glsp-client/src/browser/crossmodel-diagram-widget.tsx
ozcanxbreeze, line 1: Functionally looks good. Maybe a line of comment to describe in general what is happening in the file or what the purpose is.

Also, do you have a link to the theia events we can listen to? Just like the drop

ozcanxbreeze (changes requested):
Looks good, just need some comments at the top of the file that describe generally what the files does. This way, you wont have to look through the whole file to check what it does.

I get two bugs
- small bug, multiple instances of the same node get called 12345 instead of changing the last number, findAvailableNodeName
- unknown/Entity in the entity reference after dnd of an entity

A few questions:
- How do you bind the ctrl+space actions to the editor?
- Does this also work with outside packages? It is not in the scope of this PBI, but I saw some code coming trough for packages. I am missing some knowledge about the package handling, so I am not sure how it works yet.

extensions/crossmodel-lang/src/glsp-server/command-palette/add-entity-action-provider.ts
martin-fleck-at: Done.

extensions/crossmodel-lang/src/glsp-server/handler/add-entity-operation-handler.ts
martin-fleck-at: Done.

extensions/crossmodel-lang/src/glsp-server/handler/drop-entity-operation-handler.ts
martin-fleck-at: Done.

packages/glsp-client/src/browser/crossmodel-diagram-widget.tsx
martin-fleck-at: Done.

martin-fleck-at:
@ozcanxbreeze Thank you for the review! I added more documentation and fixed the bug in the `findAvailableNodeName` implementation. I also added test cases for it.

Regarding the other points:
- `unknown/<name>` is used when we cannot detect the necessary package information. Since by default we should always define our models in a package ("project"), I think we should keep it like that for now unless that requirement somehow changes. Please note that files in unknown packages cannot properly reference other elements so I suggest to add a `package.json` in your example workspace similar to what we have in the workspace checked into the repository.
- Ctrl+Space opens the command palette in GLSP (see registration [here](https://github.com/eclipse-glsp/glsp-client/blob/cab84ecc5a85775ceaf6c4f299e69377969b95b7/packages/client/src/features/command-palette/command-palette-tool.ts#L4)). We can add things to the command palette by implementing `ContextActionsProvider` with context id `command-palette`.
- In general we assume that all model files are defined in the scope of a package (i.e., somewhere on the file system there is a `package.json` and all files on that level or below belong to that package). If there is no package, we assign the unknown package which will have some restrictions as I mentioned above. Other than that, elements can only reference other elements within the same package unless that package has a dependency (direct or transitive) to the package where the other element is defined. The whole system is defined upon the [package.json](https://docs.npmjs.com/cli/v9/configuring-npm/package-json) format by npm so we can re-use npm to actually retrieve dependencies.

ozcanxbreeze (approved):
looks good

PR #25, opened 2023-09-21
Add menu options to add element files: entity, relationship, diagram

martin-fleck-at:
- Add entries to 'File' main menu
- Add entries to explorer context menu
- Add entries to explorer toolbar

- Re-use file name as element ID

newElements

### DIFF
--- a/packages/core/src/browser/new-element-contribution.ts
+++ b/packages/core/src/browser/new-element-contribution.ts
@@ -1,0 +1,181 @@
+/** ******************************************************************************
+ * Copyright (c) 2023 CrossBreeze.
+ ********************************************************************************/
+import { Command, CommandContribution, CommandRegistry, MaybePromise, MenuContribution, MenuModelRegistry, URI, nls } from '@theia/core';
+import { CommonMenus, DialogError, codicon, open } from '@theia/core/lib/browser';
+import { TabBarToolbarRegistry } from '@theia/core/lib/browser/shell/tab-bar-toolbar';
+import { injectable } from '@theia/core/shared/inversify';
+import { FileStat } from '@theia/filesystem/lib/common/files';
+import { FileNavigatorContribution, NavigatorContextMenu } from '@theia/navigator/lib/browser/navigator-contribution';
+import { WorkspaceCommandContribution } from '@theia/workspace/lib/browser/workspace-commands';
+import { WorkspaceInputDialog } from '@theia/workspace/lib/browser/workspace-input-dialog';
+
+const NEW_ELEMENT_NAV_MENU = [...NavigatorContextMenu.NAVIGATION, '0_new'];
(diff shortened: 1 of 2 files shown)


packages/core/src/browser/new-element-contribution.ts
```diff
@@ -0,0 +1,153 @@
...
+const NEW_ELEMENT_NAV_MENU = [...NavigatorContextMenu.NAVIGATION, '0_new'];
+const NEW_ELEMENT_MAIN_MENU = [...CommonMenus.FILE, '0_new'];
+
+interface NewElementTemplate extends Command {
+    label: string;
+    fileExtension: string;
+    content: string | ((name: string) => string);
+}
+
+const NEW_ELEMENT_TEMPLATES: NewElementTemplate[] = [
```
harmen-xb, line 22: @martin-fleck-at 

I see you add Entity, Relationship and Diagram in this order to this collection, but it doesn't appear in this order everywhere.

For the 'folder'-menu it is correct, but for the context and file menu it's not adhering the order of this array. Can you adjust it so the order is adhered?

packages/core/src/browser/new-element-contribution.ts
```diff
@@ -0,0 +1,153 @@
...
+
+    protected removeFileExtension(name: string, template: NewElementTemplate): string {
+        return name.endsWith(template.fileExtension) ? name.slice(0, -template.fileExtension.length) : name;
+    }
+}
+
+@injectable()
+export class CrossModelFileNavigatorContribution extends FileNavigatorContribution {
+    override registerCommands(registry: CommandRegistry): void {
+        super.registerCommands(registry);
```
harmen-xb, line 127: @martin-fleck-at 
If I understand this correctly this also registers the command for the command pallete?

When I type 'Entity' or 'Relationship' in the command pallete I now see the entries, which seems a bit odd. Also I think the entry should be 'New Entity' for clarity for the user.

packages/core/src/browser/new-element-contribution.ts
martin-fleck-at: Yes, I fixed the order in which the items are specified. Good catch, thank you!

packages/core/src/browser/new-element-contribution.ts
martin-fleck-at: Good point! I removed the items from the command palette (using an empty label) for now. I think that is a bit cleaner. If you want, however, we can also add them with the 'New X' label.

harmen-xb: @martin-fleck-at Please add the option in the command palette:
- New Entity
- New Relationship
- New ...

martin-fleck-at: Ok, I'll push an update as soon as https://github.com/CrossBreezeNL/crossmodel/pull/28 is merged as I have to update the PR then anyway for the initial file content.

PR #30, opened 2023-10-29
Implemented unit test in build

harmen-xb:
- Refactored github workflows to only workflow files (no actions) and include build, test and lint (including test report).
- Refactored all tsconfig files to re-use base config and all sub-folder tsconfigs to be composite.
- Fixed error in debug launch settings so debug-mode is set correctly.
- Upgraded jest dependencies to latest version.
- Refactored all jest configs to have local configs which re-use generic config.
- Performed yarn update
- Setup wokspace jest config which references to all projects.
- Updated gitpod config.
- Added VSCode Dev Container config.


harmen-xb:
@martin-fleck-at 

As part of this PR for unit tests in the build I did some extra things which I want to check with you:

1. I first manually remove some packages form the package.json file in the root. After some other commits I realized that by doing this the yarn.lock is not updated. So I performed a 'yarn upgrade' to rebuild the yarn.lock based on the specified package versions. I do realize all dependencies are also updated if newer version are available wihin the version spec. Is this ok?
2. At some point I got errors on the moduleResolution setting. I then looked into the Theia repository and sync the settings with that repo in this commit: https://github.com/CrossBreezeNL/crossmodel/pull/30/commits/13585623ac9305a241109b930e49832fab813dcf. Does this have any dowside/side-effect I am not aware of?
3. I had issues with EsLint some point which was due to an error in the include paths, was which fixed (https://github.com/CrossBreezeNL/crossmodel/pull/30/commits/0c42af9bb54620bb94ba50e92350be24cd1654ef). But before this fix I thought it was due to the tsconfig being wrong for the workspace or something. So I aligned ti to be like the Theia or Theia Blueprint repo: https://github.com/CrossBreezeNL/crossmodel/pull/30/commits/ab2a8fbb459ea2921e5154af4332a3867babd65e#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0. Is this ok? Cause to me it looks better somehow, but again I don't know the side-effect of these changes.

martin-fleck-at (changes requested):
Thank you very much for that change, @harmen-xb! Having the tests properly execute will help us a lot down the road!

As for your questions:

1. I'm not sure I fully follow you. Are you saying after moving some of the dependencies from the root package.json to other package.json files, the yarn.lock did not get updated properly? I would need to check up on this but I would not have expected this. Did you do a `yarn` or `yarn install`? Of course, you can do a `yarn upgrade` but I would be careful about it as it upgrades a lot of packages so it is better to maybe specify the particular package you want to upgrade.
2. No, I think using `Node` for module resolution should be absolutely fine. I think we switched to `Node16` as GLSP had that setting. But unless we need any specific features such as async imports it is definitely good idea to stay aligned with Theia.
3. The configuration looks good to me so thank you for fixing that! I really wouldn't expect any side effects if the linting is properly applied on all our files.
4. You are completely right, linting only does a static code analysis so there shouldn't be anything platform-specific and doing it only on Linux is definitely a good idea. I believe Theia is doing the same.

One minor thing I noticed was that when we run the linting now, we already get 3 warnings:
```
crossmodel-lang: [...]/crossmodel/extensions/crossmodel-lang/src/language-server/lexer/cross-model-indentation-tokens.ts
crossmodel-lang:   57:1  warning  Missing return type on function  @typescript-eslint/explicit-function-return-type
crossmodel-lang: ✖ 1 problem (0 errors, 1 warning)

@crossbreeze/model-service: [...]/crossmodel/packages/model-service/src/common/model-service-rpc.ts
@crossbreeze/model-service:   23:39  warning  'JsonRpcServer' is deprecated. since 1.39.0 use `RpcServer` instead  deprecation/deprecation
@crossbreeze/model-service: [...]/crossmodel/packages/model-service/src/node/model-service-backend-module.ts
@crossbreeze/model-service:   18:21  warning  'JsonRpcConnectionHandler' is deprecated. since 1.39.0 use `RpcConnectionHandler` instead  deprecation/deprecation
@crossbreeze/model-service: ✖ 2 problems (0 errors, 2 warnings)
```

Should we maybe fix them already or do you want to do this in a separate task?


Unfortunately, through the `yarn upgrade` we also did a Theia upgrade from 1.37.2 to 1.42.1 which broke the Electron build. I see the following issue when starting it up:
```
Error: Cannot find module '../src-gen/frontend/electron-main.js'
Require stack:
- [...]/crossmodel/applications/electron-app/scripts/electron-main.js
- [...]/crossmodel/node_modules/electron/dist/resources/default_app.asar/main.js
```
I believe this needs to be adapted to `lib/frontend/electron-main.js`. I didn't check for any other issues yet, not sure if you want to revert the Theia upgrade and do it separately.

martin-fleck-at (approved):
Thank you for the quick update, @harmen-xb! I had another look at it and the tests run through without any issues, all linting errors and warnings are gone and everything still works as expected. Great work!

PR #31, opened 2023-11-07
Move React components to separate package and unify data handling

martin-fleck-at:
- Extract React components to separate, Theia-less package
- Provide shared layer (reducer, context, provider) for all views
- Align usage fo ErrorView to display errors
- Improve overall typing of model reducer functionality (actions)
- Use 'immer' hooks to ease changes to immutable state handling


martin-fleck-at:
@harmen-xb I extract the React components and restructured them a bit to use a shared data layer. However, I kept the original assignment, i.e., property view vs form view. If you want, we can also align so that the form view is also using more components from the property view but I wasn't sure how much refactoring would be a good idea.

martin-fleck-at:
Re-opened as https://github.com/CrossBreezeNL/crossmodel/pull/32 to use the `feature/*` branch naming to trigger the CI.

PR #33, opened 2023-11-14
Upgrade Theia to 1.43.1 and GLSP to 2.0.0

martin-fleck-at:
Upgrade Theia to 1.43.1:
- Use Node webpack configuration for Electron for smaller package
- Adapt start script to use the new backend entry point

Upgrade GLSP to 2.0.0:
- Re-use Json-based command/operation API in GLSP server
- Adapt to new diagram configuration in GLSP client
- Remove unused CrossModelUpdateClientOperationHandler

Bonus: Fix issue with wrong updates in model service and undo/redo


harmen-xb:
I renamed the branch, so the build pipeline will execute automatically. I now see the PR is closed because of the rename.

packages/react-model-ui/package.json
```diff
@@ -38,7 +38,7 @@
     "@mui/icons-material": "^5.11.16",
     "@mui/material": "^5.13.5",
     "@mui/x-data-grid": "^6.2.1",
-    "@theia/core": "1.37.2",
+    "@theia/core": "1.43.1",
```
harmen-xb, line 41: This dependency is unexpected I think. The react component shouldn't have theia dependency right?

harmen-xb:
After running the build I see some lint issues: https://github.com/CrossBreezeNL/crossmodel/actions/runs/6868740478/job/18679997499#step:8:1

packages/react-model-ui/package.json
martin-fleck-at: Absolutely correct, I just did a search and replace but very good catch! Interestingly I thought we already got rid of it but good to fix that now, thank you!

harmen-xb: I thought so to, that's why I found it strange. But that it's gone now :)

PR #35, opened 2023-11-15
[Ignore] Testing Code QL

martin-fleck-at:


harmen-xb:
@martin-fleck-at Can you explain briefly why you did this? Do we need to do this on a regular basis?

martin-fleck-at:
I thought it would be nice to get some code analysis. We could follow up on it to detect some code smells if we want, I just started focusing on other things.

PR #38, opened 2023-11-22
Setup PlayWright tests

harmen-xb:
- Setup PlayWright test package (e2e-tests).
- Implemented CrossModel page object model (including some generic Theia ones which might be migrated to Theia in the future).
- Created 3 spec files to test basic CrossModel functionalities (to be extended in future).
- Setup execution of PlayWright and reporting in GitHub Workflow.
- Updated devcontainer configuration so PlayWright can be debugged from Windows machine (using WSL + Docker).


.github/workflows/build-and-test.yml
```diff
@@ -46,43 +46,61 @@ jobs:
...
+        uses: actions/upload-artifact@v3
+        with:
+          name: allure-results
+          path: e2e-tests/allure-results/
+          retention-days: 30
 
       # Run lint only on Linux (since it only makes sense to run it once, and linux is the fastest).
       - name: Lint
-        if: runner.os == 'Linux'
+        if: always() && runner.os == 'Linux'
```
martin-fleck-at, line 84: Could you elaborate on what is the difference between this line and just `runner.os == 'Linux'`? It seems to me that sometimes we use the combination and sometimes we do not.

.github/workflows/build-and-test.yml
```diff
@@ -46,43 +46,61 @@ jobs:
...
         with:
           name: unit-test-tesults
           # Include the unit-test-results folders (which is in the root of the workspace).
           path: unit-test-results
+          retention-days: 30
+
+      # Run PlayWright tests, only on Linux container.
+      - name: Install Playwright Browsers
+        if: runner.os == 'Linux'
+        run: yarn --cwd ./e2e-tests/ playwright install --with-deps
```
martin-fleck-at, line 68: I thin it would be nicer to put that into a script within the `package.json` on the root level or the one in the e2e-tests. What do you think?

.github/workflows/build-and-test.yml
```diff
@@ -46,43 +46,61 @@ jobs:
...
+      - name: Run Playwright tests
+        if: runner.os == 'Linux'
+        uses: coactions/setup-xvfb@v1
+        with:
+          run: yarn --cwd ./e2e-tests/ test
```
martin-fleck-at, line 73: I thin it would be nicer to put that into a script `ui-tests` or something similar within the `package.json` on the root level or the one in the e2e-tests. What do you think?

.gitignore
```diff
@@ -14,5 +14,8 @@ bin
 .ignored_out
 dist
 *.port
-**/unit-test-results
+**/*test-results
```
martin-fleck-at, line 17: I'm absolutely fine with this change but how do test results actually end up in the `test-results` folder within e2e-tests? Seems like I only get the `playwright-report`.

e2e-tests/src/page-objects/crossmodel-explorer-view.ts
```diff
@@ -0,0 +1,11 @@
+/********************************************************************************
+ * Copyright (c) 2023 CrossBreeze.
+ ********************************************************************************/
+import { TheiaExplorerView } from '@theia/playwright';
+import { TheiaTabBarToolbar } from './theia-tabbar-toolbar';
+
+export class CrossModelExplorerView extends TheiaExplorerView {
+    async tabBarToolbar(): Promise<TheiaTabBarToolbar> {
```
martin-fleck-at, line 8: It seems to me like this call wouldn't need to be async. I'd also suggest to simply create it once and store it as a property of the view to be a bit more in line with how it is done in Theia but this is really minor as the tabbar-toolbar holds no state anyway.

e2e-tests/src/page-objects/theia-single-input-dialog.ts
```diff
@@ -0,0 +1,20 @@
+/********************************************************************************
+ * Copyright (c) 2023 CrossBreeze.
+ ********************************************************************************/
+import { OSUtil, TheiaDialog, USER_KEY_TYPING_DELAY } from '@theia/playwright';
+
+export class TheiaSingleInputDialog extends TheiaDialog {
+    async enterSingleInput(inputValue: string): Promise<void> {
+        const inputField = await this.page.waitForSelector(`${this.blockSelector} .theia-input`);
+        await inputField.press(OSUtil.isMacOS ? 'Meta+a' : 'Control+a');
+        await inputField.type(inputValue, { delay: USER_KEY_TYPING_DELAY });
```
martin-fleck-at, line 10: `.type` shows up for me as deprecated and we should probably just use `.fill` here. At least for me the test is still green and I don't think we need to have a one-by-one character input here.

e2e-tests/src/tests/xmodel-app.spec.ts
```diff
@@ -0,0 +1,11 @@
+/********************************************************************************
```
martin-fleck-at, line 1: Just out of curiosity: I see that you use `xmodel` instead of `crossmodel`, is that something we should apply to more files?

martin-fleck-at (approved):
Fantastic contribution! For me locally everything works as expected and the initial CI ran with this PR also looks good. For the reporting, we need to merge to master I believe but I didn't see any obvious problems so far. I especially like the custom page objects, great work!

I do have some minor comments but they are mostly related to style.

.github/workflows/build-and-test.yml
harmen-xb: @martin-fleck-at Like discusses, the always() is needed so it also runs when any previous step has failed.

.github/workflows/build-and-test.yml
harmen-xb: I added a custom "install-playwright" script in the e2e-tests package.json for now.

.github/workflows/build-and-test.yml
harmen-xb: I added a "ui-test" yarn script in the root for this, good point!

e2e-tests/src/page-objects/crossmodel-explorer-view.ts
harmen-xb: Good point, I refacted tabBarToolbar to be a public property, which is initialized in the constructor.

Is this what you also suggest?

e2e-tests/src/page-objects/theia-single-input-dialog.ts
harmen-xb: Good point, I thought the slowed typing was nice, but it is actually slowing down the test :).

I adjusted it.

e2e-tests/src/page-objects/crossmodel-explorer-view.ts
martin-fleck-at: Yes, as a readonly public property should be fine, thank you!

e2e-tests/src/tests/xmodel-app.spec.ts
harmen-xb: That was an oopsy :). I renamed the files.

PR #40, opened 2023-11-24
Upgrade to Langium 2.1.3

martin-fleck-at:
Langium > 2.0.0 only supports ESM so we refactor the extension to ESM

Setup Changes
- Set type to 'module' in package.json and update configs accordingly
- Use esbuild to develop in ESM but compile to CJS for VSX

Code Changes
- Adapt all relative imports to use '.js' extension (even for '.ts')
- Remove now obsolete adaptations of workspace manager

Test Changes
- Provide custom Jest configuration for ES modules
- Use 'jest.useFakeTimers()' as specified in Jest documentation for ESM
- Add '--experimental-vm-modules' Node flag to ESM test executions
- Add 'NODE_NO_WARNINGS' environment variable to ignore ESM warning
- Use unique report file names as we need to run twice (CJS, ESM)
- Add display name to all test configs for nicer output
- Adapt launch configs to properly debug single test or all tests


extensions/crossmodel-lang/src/glsp-server/handler/add-entity-operation-handler.ts
```diff
@@ -5,19 +5,19 @@
...
+import { CrossModelCommand } from './cross-model-command.js';
 
 /**
  * An operation handler for the 'AddEntityOperation' that resolves the referenced entity by name and places it in a new node on the diagram.
  */
 @injectable()
 export class CrossModelAddEntityOperationHandler extends JsonOperationHandler {
     override operationType = AddEntityOperation.KIND;
-    @inject(ModelState) protected override modelState: CrossModelState;
+    @inject(ModelState) protected override modelState!: CrossModelState;
```
harmen-xb, line 20: @martin-fleck-at Out of curiousity, what does the ! do? Is it enforcing non-nullability or something?

extensions/crossmodel-lang/src/extension.ts
```diff
@@ -35,7 +34,7 @@ function launchLanguageClient(context: vscode.ExtensionContext): LanguageClient
 
 function createServerOptions(context: vscode.ExtensionContext): ServerOptions {
    // needs to match the configuration in tsconfig.json and webpack.config.js
-   const serverModule = context.asAbsolutePath(path.join('out', 'server-main'));
+   const serverModule = context.asAbsolutePath(path.join('out', 'main.cjs'));
```
harmen-xb, line 37: So you can mix .js and .cjs imports? cjs is for the CommonJS files? as opposed to .js files now being ESM?

extensions/crossmodel-lang/package.json
```diff
@@ -94,27 +97,29 @@
     "@eclipse-glsp/layout-elk": "2.0.0",
     "@eclipse-glsp/server": "2.0.0",
     "chalk": "~4.1.2",
-    "chevrotain": "~10.4.2",
+    "chevrotain": "~11.0.3",
```
harmen-xb, line 100: @martin-fleck-at 

When @ozcanxbreeze worked on the Langium implementation he ran into some issue related to this PR in chevrotain: https://github.com/Chevrotain/chevrotain/pull/1917.

Because of this reason we didn't use a higher chevrotain version. Is this problem not occuring with the latest chevrotain (since I see you upgraded it?

extensions/crossmodel-lang/test/language-server/lexer/cross-model-lexer.test.ts
```diff
@@ -2,14 +2,16 @@
...
-import { describe, expect, test, beforeAll } from '@jest/globals';
+import { describe, expect, test, beforeAll, jest } from '@jest/globals';
 import { EmptyFileSystem } from 'langium';
 import { tokenMatcher } from 'chevrotain';
 
 import { CrossModelLexer } from '../../../src/language-server/lexer/cross-model-lexer';
 import { DEDENT, INDENT } from '../../../src/language-server/lexer/cross-model-indentation-tokens';
 import { createCrossModelServices } from '../../../src/language-server/cross-model-module';
 
+jest.useFakeTimers();
```
harmen-xb, line 13: Why is this needed?

harmen-xb (commented):
All in all looks good to me, just some small questions in comments.

extensions/crossmodel-lang/src/glsp-server/handler/add-entity-operation-handler.ts
martin-fleck-at: Yes, exactly. I'm not sure why we but we do get a warning now that the properties are not initialized (directly or in the constructor) because they are injected. So in this case, I assert the non-nullability with `!` to avoid that warning.

extensions/crossmodel-lang/src/extension.ts
martin-fleck-at: Yes, we use `.cjs` to denote CommonJS files and assume that all other files are ESM in the extension. However, VS Code extensions only work with CommonJS code so we transpile the two necessary extension files (`extension.ts` and `main.ts`) as CommonJS in the builder (`esbuild.mjs`).

configs/jest.config.js
```diff
@@ -1,5 +1,5 @@
+/** @type {import('ts-jest').JestConfigWithTsJest} */
 module.exports = {
     rootDir: '../',
-    projects: ['<rootDir>/applications/*', '<rootDir>/extensions/*', '<rootDir>/packages/*'],
-    reporters: ['default', ['jest-junit', { outputDirectory: 'unit-test-results', outputName: 'jest-report.xml' }]]
```
harmen-xb, line 4: This change is leading to the Jest extension in VSCode not to work anymore. I am trying to fix this, but I guess Jest needs to run either for ModuleJS or ESM?

extensions/crossmodel-lang/package.json
martin-fleck-at: Yes that was fixed in https://github.com/Chevrotain/chevrotain/commit/d50420dc64e52310250136dc9e848057f2e8afb7, so very recently!

extensions/crossmodel-lang/test/language-server/lexer/cross-model-lexer.test.ts
martin-fleck-at: I cannot say for sure but it was mentioned in the [Jest documentation for using ESM](https://jestjs.io/docs/ecmascript-modules#differences-between-esm-and-commonjs).

harmen-xb: I think the code is just an example of using the jest import.

If I read here: https://jestjs.io/docs/timer-mocks#enable-fake-timers I think it's only needed of you need to mock timers.

martin-fleck-at: I had some test failures when I removed it in one of the tests so I added it everywhere but I can have a closer look again.

martin-fleck-at: Yes, I get the following error in every single test file where I do not have the statement (I left it in one that is why 1 passes):
```
crossmodel-lang: npm ERR! Lifecycle script `test` failed with error: 
crossmodel-lang: npm ERR! Error: command failed 
crossmodel-lang: npm ERR!   in workspace: crossmodel-lang@0.0.1 
crossmodel-lang: npm ERR!   at location: /[...]/crossmodel/extensions/crossmodel-lang 
lerna ERR! execute callback with error
lerna ERR! Error: Command failed: npm run test
lerna ERR! Test Suites: 9 failed, 1 passed, 10 total
lerna ERR! Tests:       6 passed, 6 total
lerna ERR! Snapshots:   0 total
lerna ERR! Time:        11.764 s
lerna ERR! Ran all test suites.
lerna ERR! npm ERR! Lifecycle script `test` failed with error: 
lerna ERR! npm ERR! Error: command failed 
lerna ERR! npm ERR!   in workspace: crossmodel-lang@0.0.1 
lerna ERR! npm ERR!   at location: /[...]/crossmodel/extensions/crossmodel-lang 
lerna ERR! 
lerna ERR! 
lerna ERR! > crossmodel-lang@0.0.1 test
lerna ERR! > cross-env NODE_OPTIONS=--experimental-vm-modules NODE_NO_WARNINGS=1 jest --passWithNoTests
lerna ERR! 
lerna ERR! 
lerna ERR!     at /[...]/crossmodel/node_modules/lerna/node_modules/execa/index.js:236:11
node:internal/process/promises:289
            triggerUncaughtException(err, true /* fromPromise */);
            ^
```

Do you see the same? I can't find any more documentation and the generic one (https://jestjs.io/docs/jest-object#fake-timers) does not give any hints why we need it with ESM.

configs/jest.config.js
harmen-xb: The Jest extension in VSCode is using the "jest.jestCommandLine" setting to execute tests (https://github.com/CrossBreezeNL/crossmodel/blob/b94137d75b41879c858d32a9e5c5b75ffaaf7a92/.vscode/settings.json#L43C3-L43C77).

I tried to change it to the setting for ESM for testing the extension (since this is the only package which has unit tests a.t.m. I tried (amonst other things):

"jest.jestCommandLine": "npm exec -- cross-env NODE_OPTIONS=--experimental-vm-modules NODE_NO_WARNINGS=1 jest --config=./configs/esm.jest.config.js"

But while this works fine on the command line (also with --watch and --watchAll) it doesn't work with the extension.

Do you have any idea?

harmen-xb: I think without changing anything the tests work in all other projects (except the extension). But once I change the command to execute the extension jest stuff it doesn't work.

Maybe for now it's acceptable, I just need to remember to run the extension tests seperately.

extensions/crossmodel-lang/test/language-server/lexer/cross-model-lexer.test.ts
harmen-xb: @martin-fleck-at On my machine it complained about the imports, so I added the .js at the end first. Then removed the jest.useFakeTimers(); call and it works on my machine.

With only removing the jest.useFakeTimers(); it indeed gives errors.

martin-fleck-at: Interesting but you are right! Thank you!

configs/jest.config.js
martin-fleck-at: I adapted everything a bit and the extension works better now. However I didn't get the debugging support to work properly in the Test view.

harmen-xb: Debugging I can do with de VS Code launch. I like the extension because it run the test automatically when you change a file.

It works now also within my setup, great!

harmen-xb (approved):
All looks good now!

Issue #36, opened 2023-11-15
ERROR Error: Cannot update an unmounted root.


harmen-xb:
Sometimes I get the following error:
```
2023-11-15T10:23:15.321Z root ERROR Error: Cannot update an unmounted root.
    at ReactDOMHydrationRoot.render.ReactDOMRoot.render (http://localhost:3000/bundle.js:145610:11)
    at FormEditorWidget.onUpdateRequest (http://localhost:3000/bundle.js:62984:23)
    at __webpack_modules__.../../node_modules/@phosphor/widgets/lib/widget.js.Widget.processMessage (http://localhost:3000/bundle.js:161321:22)
    at invokeHandler (http://localhost:3000/bundle.js:16566:21)
    at sendMessage (http://localhost:3000/bundle.js:16302:13)
    at runMessageLoop (http://localhost:3000/bundle.js:16616:17)
```

**Steps to reproduce**
1. Start CrossModel with yarn start:browser
2. Open the yaml-example workspace
3. Open the form-editor on the customer entity.
4. Close the form-editor on the customer entity.
5. Open the diagram
6. Change the customer entity via the property widget (for example change the name and press 'Save').

After step 6 the exception will be in the console.

If you don't do step 4, the exception doesn't occur. So it looks like the form-editor-widget is not unregistering itself from the model-server.

PR #43, opened 2023-11-29
Ensure form client's update listener is disposed properly

martin-fleck-at:
Fixes https://github.com/CrossBreezeNL/crossmodel/issues/36


harmen-xb (approved):
Nice one :).

Issue #16, opened 2023-05-04
Can't have multiple diagrams open at the same time


ozcanxbreeze:
When you open the graphical view of two different models, the application freezes most of the time.

willem-xbreeze:
Lookin into the issue:
When two diagram views are opened within either the same model or two different models, the second one does not show all objects, but when looking in the DOM they appear to be there

harmen-xb:
**Steps to reproduce:**

1. Start CrossModel with yarn start:browser
2. Open the yarn-example workspace
3. Create a new diagram and drag some entities in them.
4. Open a second diagram (the existing one).

At step 4 the diagram seems empty.

Now if you close both diagrams and open one of them you see some zooming happening. So it looks like there is something off with the zoom of the diagrams.

PR #44, opened 2023-11-30
Allow multiple diagrams to be opened

martin-fleck-at:
- Fix TheiaJsonrpcGLSPClient to only have one handler for connection
- Fix problem with filter with same ID being added multiple times

- Add formatting of CSS files to prettier
- Centralize ID handling for containers and global IDs -- Diagram elements need locally unique IDs
-- Other elements need globally unique IDs

Fixes https://github.com/CrossBreezeNL/crossmodel/issues/16


harmen-xb (changes requested):
Hi Martin,

Thanks for these changes. Codewise I don't see strange things.

I do see the PlayWright tests timed-out, so I updated the playwright config so it will really error on actions which take to long (like waitFor* actions). Otherwise it might end-up unnoticed.

While testing I experiences some issues, so I list them below here:

## 2. Diagram context menu issue
I also have another issue with the context menu re-occurring.
Sometimes when I right-click a diagram node in the explorer view it just misses a lot of option. See for example here:
image

If I then click on another node, and back and then right-click it doesn work agaain:
image

We had this issue before on entities, but now I can also reproduce it on diagram files (which seems to be related to a diagram being opened).

What I also noticed that when the context menu is wrong the keyboard actions also don't work. Like deleting the file with the 'Delete' key doesn't work, this is also not in the context menu if it's wrong. So I guess this is linked

Is this a known Theia/GLSP issue?

### Steps to reproduce:

1. Open the yaml-example workspace
2. Make sure no diagram is open
3. Click on a diagram so it opens
4. Right-click on the diagram file to open the context menu.

It this point the context menu doesn't contain all options. When I choose the 'Copy' option and paste it in a text editor I get the full path of the diagram. So it does seem to execute the action on the right node.

Once you have the diagram open, and leave it open the problem doesn't re-occur when working with different files and jumping back. So it seems related to the diagram tab being opened.

harmen-xb:
## Issue: Couldn't find node
Also I experience an error when modeling a new diagram, where I get an error that it couldn't find a node with the given id:
```
2023-11-30T21:57:46.157Z root ERROR Request requestDiagramNodeEntityModel failed with error: Request server/requestModelDiagramNode failed with message: No node found with the given id: CustomerToOrder Error: Request server/requestModelDiagramNode failed with message: No node found with the given id: CustomerToOrder
    at handleResponse (/workspaces/crossmodel/node_modules/vscode-jsonrpc/lib/common/connection.js:565:48)
    at handleMessage (/workspaces/crossmodel/node_modules/vscode-jsonrpc/lib/common/connection.js:345:13)
    at processMessageQueue (/workspaces/crossmodel/node_modules/vscode-jsonrpc/lib/common/connection.js:362:17)
    at Immediate.<anonymous> (/workspaces/crossmodel/node_modules/vscode-jsonrpc/lib/common/connection.js:334:13)
    at processImmediate (node:internal/timers:466:21)
```

### Steps to reproduce:

1. Open the yaml-example workspace
2. Create a new diagram
3. Drag an the two existing entities.
4. Create a new relationship using the GLSP Toolbox between the two entities.
5. Now click on the relationship in the diagram, then the error is in the terminal.

When I look in the diagram code I see the following contents:

```
    edges:
      - id: "CustomerToOrder"
        relationship: "unknown/CustomerToOrder"
        sourceNode: "CustomerNode"
        targetNode: "OrderNode"
```

When I compare this with the other diagram which is working, the `unknown/` seems to be the issue. So it looks like local ids are not created correctly yet. When I look at the contents of the new relationship I also see the two entities are references prefixed with `unknown/`.

harmen-xb:
## Issue: Diagram context menu issue
I also have another issue with the context menu re-occurring.
Sometimes when I right-click a diagram node in the explorer view it just misses a lot of option. See for example here:
image

If I then click on another node, and back and then right-click it doesn work agaain:
image

We had this issue before on entities, but now I can also reproduce it on diagram files (which seems to be related to a diagram being opened).

What I also noticed that when the context menu is wrong the keyboard actions also don't work. Like deleting the file with the 'Delete' key doesn't work, this is also not in the context menu if it's wrong. So I guess this is linked

Is this a known Theia/GLSP issue?

### Steps to reproduce:

1. Open the yaml-example workspace
2. Make sure no diagram is open
3. Click on a diagram so it opens
4. Right-click on the diagram file to open the context menu.

It this point the context menu doesn't contain all options. When I choose the 'Copy' option and paste it in a text editor I get the full path of the diagram. So it does seem to execute the action on the right node.

Once you have the diagram open, and leave it open the problem doesn't re-occur when working with different files and jumping back. So it seems related to the diagram tab being opened.

harmen-xb:
## Issue: Property widget starts flashing (yes you read it correctly 😄 )

At some point I saw the property widget started flashing, looks like it is re-rendering continuously.

### Steps to reproduce:

1. Open the yaml-example workspace
2. Make sure the property view is visible
3. Open the existing diagram
4. Make a new diagram and drag the Customer entity into it.
5. Click on the Customer entity

At this point on Chrome Windows the property widgets kind of starts flashing, where the contents looks to be reloaded every second.

Once you go to the other diagram and click on an entity it works fine. If you then jump back to the new diagram and click on an entity it also works fine.

martin-fleck-at:
> requestDiagramNodeEntityModel

Yes, I think that probably never worked. I just checked the code and it comes from the property view that requests an entity but when you click on an edge the server rightfully says that there is no **node** with the given ID because we clicked on an edge. I'll see what we can do to best prevent that.

harmen-xb:
Hi @martin-fleck-at,

I see you pushed some changes to fix the issues.

I tested it and all issues indeed seem to be solved.

A new issue I now see is that the property widget form is working (it when I press Save the entity file is updated), but then the property widget is emptied and the diagram is not updated. So something is not OK yet.

martin-fleck-at:
@harmen-xb Interesting, I'll double check that, thank you for the quick review!

martin-fleck-at:
@harmen-xb I pushed another update that should fix that issue. Could you please have another look?

harmen-xb:
@martin-fleck-at Thanks for the update! It now all seems the work correctly.

To make sure the context menu's keep working I added some extra playwright tests on the context menu of relationship and diagrams.

PR #46, opened 2024-02-08
Provide diagram support for mapping files

martin-fleck-at:
- Move and rename previous diagram classes to 'System Diagram'
- Introduce new 'Mapping Diagram' contribution in GLSP (server, client)
- Support dropping entities on diagram for new source objects
- Support command palette (Ctrl+Space) to create new source objects
- Support create attribute mappings by dragging attributes
- Support create attribute mappings by specifying literals on canvas
- Support deletion of elements and mappings

- Use auto-layout for diagram (source objects left, target right)
-- Do not create dedicated diagram file with layout information
-- Auto layout elements manually until Elk can be properly used

- Adapt test cases

- Add handler for unhandledPromise rejections to avoid server crashes

- Improve styling for all diagrams
-- Provide grid movement
-- Provide proper background
-- Improve styling of tool palette
-- Color node types

mapping-diagram


harmen-xb:
Hi @martin-fleck-at,

Thanks for the update. The screens look very promising!

I will try to make time tomorrow to do a proper review.

harmen-xb:
Hi @martin-fleck-at,

I am trying to test the new mapping diagram, but I keep getting the following error when opening the existing or a new mapping diagram:

image

Details:
image

This error is only popping up in the UI, it's not in the browser console (or terminal).

martin-fleck-at:
@harmen-xb No, the yarn.lock file should be stable. I'll try again maybe it is the different node versions, I think I set my default to 20 but I see that we are using Node 16 in the dev container so that is probably why some functions are not available. I'll have a look and try with Node 16, sorry for that!

martin-fleck-at:
@harmen-xb I pushed an update for Node 16-compatibility. We can definitely use Node 20 as Theia is also built against it (as well as Node 16 and Node 18, see https://github.com/eclipse-theia/theia/blob/master/.github/workflows/ci-cd.yml#L55) but maybe we should open a separate PR for it and see if there are more places where we could update some code and we also would want to update the Dockerfile.

harmen-xb:
Hi @martin-fleck-at,

Thanks for the update, I built it again with node16 in the DevContainer and now it indeed works. It seems to work quite well with dragging, and auto-layout.

I do have some remarks:
- When I open the mapping diagram and then add an attribute to a target or source entity, the diagram is not updated automatically.
- I see you committed a fix for the tool palette of literal, but I can't find any such functionality. Is there an option to create a literal in the diagram somehow?

harmen-xb:
@martin-fleck-at Thanks for the response. For now the literal functionality is quite good I think.

There are some small things, like I can map to a target attribute multiple times now, the grammer also accepts this. This is something we will think about how this should be in the future.

For now this is definitely good enough as a first version.

harmen-xb (approved):
Code looks good to me.

PR #54, opened 2024-04-10
Introduce form to edit Relationship data and attributes

martin-fleck-at:
Introduce reference resolution and completion mechanism for elements
- Ensure we send the global id as part of each identifiable element
- Attach reference data to GLSP elements for property view
- Replace specific 'requestDiagramNodeEntityModel' method
- Replace specific 'findRootReferenceName' method

Create Form for Relationship data
- Unify form for property view and editor (also for Entity)
- Use MUI components consistently and add theming for them
- Remove custom stylesheets
- Remove dependency to 'react-tabs' as it is no longer used

Refactorings:
- Rename 'ExternalId' to 'GlobalId'
- Ensure IDs do not contain any invalid characters
- Fix issue with attribute serialization for relationships


martin-fleck-at:
@harmen-xb The PR is ready for review, the Mac build keeps failing for some reason, I'll try to re-trigger it a bit later, maybe it is a temporary problem.

harmen-xb:
> @harmen-xb The PR is ready for review, the Mac build keeps failing for some reason, I'll try to re-trigger it a bit later, maybe it is a temporary problem.

Hi @martin-fleck-at ,

The mac build reran a couple of times but it keeps getting out of memory: https://github.com/CrossBreezeNL/crossmodel/actions/runs/8644531235/job/23747744720#step:5:253

I re-ran the workflow on the main branch to check whether it is a container issue, but it succeeds there. So there seems to be some changes which causes this.

martin-fleck-at:
@harmen-xb Interesting! Thank you for following up. I'll see if there is anything I can do to find what would be causing this.

harmen-xb:
It looks like the issue is a quite common one when running out of memory: https://medium.com/@nirbhaysingh281/fatal-error-ineffective-mark-compacts-near-heap-limit-allocation-failed-javascript-heap-out-of-5637924b466

martin-fleck-at:
@harmen-xb Unfortunately my idea to upgrade to Node 20 caused some issues on the Windows build with `node-pty`. I didn't want to invest too much time, so I just upped the Node memory for the build job to 8GB ([specs here](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories)) which seemed to work. If we want to upgrade Node, we should do it separately then.

packages/react-model-ui/package.json
```diff
@@ -35,12 +35,12 @@
     "@crossbreeze/protocol": "0.0.0",
     "@emotion/react": "^11.10.6",
     "@emotion/styled": "^11.10.6",
-    "@mui/icons-material": "^5.11.16",
-    "@mui/material": "^5.13.5",
-    "@mui/x-data-grid": "^6.2.1",
+    "@mui/icons-material": "^5.15.15",
+    "@mui/lab": "5.0.0-alpha.170",
```
harmen-xb, line 39: @martin-fleck-at I don't see this dependency being used anywhere.

harmen-xb (commented):
One remark on the code.

packages/react-model-ui/package.json
martin-fleck-at: @harmen-xb Good catch! I forgot to remove it after a few rounds of iterations. I can do it as soon as your build runs through again.

harmen-xb:
@martin-fleck-at

Couple of things:

1. I notice when I edit any relationship and don't press save, click away and click on the edge again the change is gone. Which kinda makes sense, but it's not warning me about losing changes or anything. Would it be easy to add something like it?
2. CTRL+S is not working in the property widget. Would it be possible to enable the keybinding here?
3. When I add a child and parent attribute for a relationship and press 'Save' it's not persisted. I also see an error on the console:
```
2024-04-16T10:23:03.354Z root WARN MUI: The value provided to Autocomplete is invalid.
None of the options match with `{"uri":"","label":"ExampleCRM.Address.Street","type":"Entity"}`.
You can use the `isOptionEqualToValue` prop to customize the equality test.
2024-04-16T10:23:05.927Z root ERROR Request update failed with error: Request server/update failed with message: invalid Error: Request server/update failed with message: invalid
    at handleResponse (/workspaces/crossmodel/node_modules/vscode-jsonrpc/lib/common/connection.js:565:48)
    at handleMessage (/workspaces/crossmodel/node_modules/vscode-jsonrpc/lib/common/connection.js:345:13)
    at processMessageQueue (/workspaces/crossmodel/node_modules/vscode-jsonrpc/lib/common/connection.js:362:17)
    at Immediate.<anonymous> (/workspaces/crossmodel/node_modules/vscode-jsonrpc/lib/common/connection.js:334:13)
    at processImmediate (node:internal/timers:466:21)
```
4. When reordering attributes of an entity or attribute-pairs in a relationship the icon for moving attribute up for the first item and moving down for the last item would be better to disable it (or hide it)

harmen-xb:
@martin-fleck-at

It all works nicely now. I did notice one strange thing:
- When in a system diagram and you click on a relationship or entity with a Description and then click on any object of the same type without a Description the value of the previous object is still there while all other properties are updated. When switching object type the Description is updated, so there seems to be something specific when switching between properties of the same object type.

martin-fleck-at:
@harmen-xb Very good catch! It seems that MUI doesn't like undefined values but when rendering empty strings everything works as expected.

harmen-xb (approved):
All looks ok now

PR #55, opened 2024-05-17
Add import/export functionality for a single system/package

martin-fleck-at:
- Model files and the 'package.json' are stored in a single YAML file
-- The official YAML document delimiter is used between documents
-- We store some additional metadata as comments for now

- Import always creates a folder on workspace root-level
-- User is asked to overwrite if folder already exists


harmen-xb:
@martin-fleck-at I reviewed the export functionality and I noticed a couple of things:

1. I expected the "Export System" would work on the System folder. But it seems it only works on any file inside the system, folder. This is a bit counter intiutive. I would expect when I would right click on the 'ExampleDWH' folder and click 'Export System' it would work, but I get an error: 
image. So for now I think it's best to change the menu item so it only appears when it is a system folder.
2. When I export any system from a file inside the system a dialog box appears which shows me the contents of the export. By default the Format is set to 'System Export' which only shows folders (or is it the filter for cmexport files?). I can select 'All Files', which then shows all the files. In this screen it wasn't immediatly clear to me that I need to specify the export target. I thought I could select what to export or something. So I think there needs to be a short explanation in the dialog screen. Something like 'Please choose the location and enter the file name for the export file.'.
3. In the export file dialog, the title doesn't need a file counter, 'Export ExampleDWH' is enough.
4. I think the import in the context menu of the explorer is a bit confusing. I noticed when you right-click on an export file and choose import it immediately imports the system into the root of the workspace. But if you click enywhere else you get a dialog box to choose an import file and then it also imports it into the root of the workspace. Changes required:
   - For the second dialog box I think we need to add some explanation there as well. Something like 'Please select the export file to import.'.
   -    To make it consistent I think it would be nicer if you can only choose 'Import System' on a folder or on a '*.cmexport' file. When clicking on a *.cmexport file I think it should ask for the target location for the import. And when choosing "Import System" on a folder it should import the model into the selected folder.

martin-fleck-at:
@harmen-xb I tried to address all your points in a follow-up commit. Adding messages to the file dialog is not that easy but I adapted the title a bit and with the new, more restricted, workflow I believe the flow is much better and it is much clearer what needs to be done. It would be really great if you could have another look. Thank you!

harmen-xb:
@martin-fleck-at 

Thanks for the update.

Some feedback:

1. In the context menu for export it states "Export System as File", this should be "Export System to File"
2. The file format selector is now on "All Files" by default, I think having the filter on "System Export Files" makes more sense in the new setup. (for import and export)
3. Can you change the location the export file explorer opens from the system folder to workspace folder? It unlikely a user want/should export to the folder it is exporting.
4. The context menu on the .cmexport file for importing the system is gone. Is this on purpose?

martin-fleck-at:
> 1. In the context menu for export it states "Export System as File", this should be "Export System to File"

Fixed in the new commit.

> 2. The file format selector is now on "All Files" by default, I think having the filter on "System Export Files" makes more sense in the new setup. (for import and export)

Fixed in the new commit.

> 3. Can you change the location the export file explorer opens from the system folder to workspace folder? It unlikely a user want/should export to the folder it is exporting.

Fixed in the new commit.

> 4. The context menu on the .cmexport file for importing the system is gone. Is this on purpose?

Yes that was on purpose since the entry point for the import is now a directory to which you want to import and you select the file afterwards. However, if you want, we can add a second entry point that if an export file is selected then you need to choose the directory into which you want to import. I just thought it might be a bit confusing but maybe not.

harmen-xb:
@martin-fleck-at 

Thanks for the update, sorry for the delay in response. I missed the notification of you comment.

All looks good now, one small think. Is it possible to move the file extension selector to the right or below the file name in the file selector?
image

At least on Windows it's never before/above.

harmen-xb (approved):
Code changes look good,.

PR #56, opened 2024-05-22
Add support for textual expressions in attribute mappings

martin-fleck-at:
- Allow specifying expressions in attribute mappings as string
- Validate reference expressions '{{expression}}'
- Provide suggestions for reference expressions
- Allow attribute mappings to have 0 to n sources (previously 1)

Minor:
- Guard return value in selection data service for undefined references

### DIFF
--- a/examples/mapping-example/ExampleDWH/CompleteCustomerMapping.mapping.cm
+++ b/examples/mapping-example/ExampleDWH/CompleteCustomerMapping.mapping.cm
@@ -15,14 +15,22 @@ mapping:
         entity: ExampleDWH.CompleteCustomer
         mappings:
           - attribute: Country
-            source: CustomerSourceObject.Country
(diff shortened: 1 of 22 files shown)


harmen-xb:
@martin-fleck-at 

Expressions look good and nice to see the completion working!

Some things:
- The suggestion in the expression works when we type {{ and press ctrl+space. But when we have {{ with any character after it, like a partial name, it doesn't work.
- When clicking on a validation error ("Only sources can be ...") it opens  the graphical mapping, i.s.o. jumping to the code where the error is.

martin-fleck-at:
@harmen-xb I adapted the behavior as you mentioned and now everything should work as expected. Please have another look.

harmen-xb (approved):
All works good now, code also looks good as far as I can check.

PR #57, opened 2024-06-06
Add property view editor for attribute mapping

martin-fleck-at:
- Remove string and number literals, use expressions instead
- Mark target attributes with an expression in diagram
- Add form for attribute mapping in properties view
- Generalize AttributeGrid to GridComponent


harmen-xb:
@martin-fleck-at  Is this ready to be reviewed?

martin-fleck-at:
@harmen-xb Yes, sorry, forgot to add you as a reviewer as I was waiting for the CI results.

harmen-xb:
@martin-fleck-at 

Nice to see the property widget for the attribute mappings, this is very nice addition!

Some feedback:
- The fx behind the attribute, is it easy to make this another color so it's more clear it's not part of the attribute name? (maybe bit greyed out)
- The expression editor in the widget doesn't have auto completion. Is it hard to add this logic here as well?
- On 'Sources' when you click 'Add source' it's not very clear you can click/start typing in the new row. Is it possible to give the cell an instruction in grey or something that it's more clear. Something like "Select an attribute.."?
- When you remove a attribute mapping by clicking on a line and pressing delete, the expression of all attribute mappings are gone. I would expect only the source attribute to be deleted.
- The property widget doesn't show the form for attributes which don't have a source or expression. I think it should always show the form (also if there is no attribute mapping yet).

martin-fleck-at:
> * The fx behind the attribute, is it easy to make this another color so it's more clear it's not part of the attribute name? (maybe bit greyed out)

Done, it is now rendered in a grey color.

> * The expression editor in the widget doesn't have auto completion. Is it hard to add this logic here as well?

Adding that to the form is quite a bit more work and I suggest to do it separately since the form is not equivalent to the textual representation in other aspects as well, e.g., validation.

> * On 'Sources' when you click 'Add source' it's not very clear you can click/start typing in the new row. Is it possible to give the cell an instruction in grey or something that it's more clear. Something like "Select an attribute.."?

I improved the logic so that we get proper auto-focus on the text field and show a placeholder text like you suggested.

> * When you remove an attribute mapping by clicking on a line and pressing delete, the expression of all attribute mappings are gone. I would expect only the source attribute to be deleted.

That was a mistake indeed. I deleted all mappings that didn't have any sources but "empty" mappings are allowed.

> * The property widget doesn't show the form for attributes which don't have a source or expression. I think it should always show the form (also if there is no attribute mapping yet).

Since we allow empty mappings, I now initialize mappings with an empty attribute mapping for each target attribute. This can serve as a useful starting point for the user in the textual editor but also allows easier logic in other editors.

harmen-xb (approved):
Code looks good to me.

PR #58, opened 2024-06-15
Support editing source object specification in property view

martin-fleck-at:
- Rename 'relations' to 'dependencies'
- Remove relationship conditions and only keep join conditions (MVP)
- Enhance expression to support more operators and literals
- Add form for source objects to edit dependencies and join condition
- Improve completion provider for non-text keywords
- Split model reducer actions into separate files for better devex

### DIFF
--- a/examples/mapping-example/ExampleDWH/mappings/CompleteCustomerMapping.mapping.cm
+++ b/examples/mapping-example/ExampleDWH/mappings/CompleteCustomerMapping.mapping.cm
@@ -7,10 +7,10 @@ mapping:
       - id: CalcAgeSourceObject
         entity: ExampleDWH.CalcAge
         join: apply
-        relations:
+        dependencies:
           - source: CustomerSourceObject
             conditions:
-              - join: CalcAgeSourceObject.BirthDate = CustomerSourceObject.BirthDate
+              - CalcAgeSourceObject.BirthDate = CustomerSourceObject.BirthDate
     target:
         entity: ExampleDWH.CompleteCustomer
         mappings:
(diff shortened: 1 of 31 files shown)


harmen-xb:
@martin-fleck-at 

After testing the new source object setup we stumbled across a limitation which needs to be addressed. I wrote the changes in the user story here: https://dev.azure.com/x-breeze/CrossModel/_workitems/edit/869

martin-fleck-at:
@harmen-xb Thank you for the feedback! I'll try to push an update by end of day tomorrow.

martin-fleck-at:
@harmen-xb I pushed another commit that should take of moving the conditions up to the source object directly.

harmen-xb:
@martin-fleck-at 

The changes look good. This definitely makes sense modeling wise.

I have a couple of change requests:

1. I see an issue with the conditions when the source object is removed. If you create conditions containing a source object, these are not removed when you remove the source object. At a minimum I think we should validate the conditions, that they refer only to source objects which are listed in the dependencies (or the current object or literals).
1. Another issue I see is that you can select the current object as a dependency. I think you should remove the current object from the drop-down to make sure this is not selected. Also a validation needs to be added that a dependency can only be an source object within the same mapping, except for the current source object.
1. Could you add another validation: within one mapping there can only be one sourceobject with jointype from.
1. In the conditions part of the property form, could you make the Operator column written in full i.s.o. the 'OP'? I think there is enough space and makes it a lot more clear.

martin-fleck-at:
@harmen-xb Thank you for your feedback! I pushed another commit that should now hopefully iron out the last issues.

harmen-xb (approved):
Changed look good to me.

PR #59, opened 2024-06-27
Add identifier property to Entity attributes

martin-fleck-at:
- Highlight identifier attributes by underlining their name
- Allow to set identifier property in Entity form

### DIFF
--- a/examples/mapping-example/Sources/ExampleCRM/entities/Address.entity.cm
+++ b/examples/mapping-example/Sources/ExampleCRM/entities/Address.entity.cm
@@ -3,10 +3,10 @@ entity:
     name: "Address"
     description: "The address of a customer."
     attributes:
-      - identifier: true
-        id: CustomerID
+      - id: CustomerID
         name: "CustomerID"
         datatype: "Integer"
+        identifier: true
       - id: Street
         name: "Street"
         datatype: "Text"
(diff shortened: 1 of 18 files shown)


harmen-xb:
@martin-fleck-at 

Nice work.

Some remarks:
- Can you change the order of the properties in the grammer of an attribute so the 'identifier' property is listed after the datatype? Same for the form, it can be positioned after the Data Type property.
- Is it possible in the form to change the 'ID' label to a key-icon? Like this one: https://fonts.google.com/icons?selected=Material+Symbols+Outlined:key_vertical:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=key&icon.size=24&icon.color=%235f6368.
- And in the form change the appearance of the values so ticked is https://fonts.google.com/icons?selected=Material+Symbols+Outlined:check_box:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=tick&icon.size=24&icon.color=%235f6368 and not-ticked is https://fonts.google.com/icons?selected=Material+Symbols+Outlined:check_box_outline_blank:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=tick&icon.size=24&icon.color=%235f6368? And then when you double click you can actually edit it using the normal tick-box input (like it is currently).

martin-fleck-at:
@harmen-xb Sure, I pushed an update that should make all the changes that you requests.

harmen-xb (approved):
Code looks good

harmen-xb:
@martin-fleck-at One last question to discuss in 5 minutes, is it simple to change the structure diagram to show the key icon in front of key attributes i.s.o. the underline?

martin-fleck-at:
I pushed an update that changes the underline for a key icon.

harmen-xb:
Yes, that looks way better :). Thanks!

PR #60, opened 2024-07-05
Upgrade Frameworks and Improve Tools and Toolbox in Diagrams

martin-fleck-at:
Upgrade Frameworks:
- Update to Theia 1.49.1 - the latest community release
- Update to GLSP 2.2.0 - the latest GLSP version
- Update to Node 20 - only affects CI workflows

Improve Tools and Toolbox in Diagrams:
- Wrap tool functionality up into their own tool modules
- Define default tool as select tool
- Create new tool for creating an entity in the system diagram
- Ensure newly created files are put into the correct sub-folder
- Render toolbox to explicitly list tools and give toolbox better style


martin-fleck-at:
@harmen-xb I have no idea why the Windows build keeps failing. Could you please check if that also happens locally on your Windows machine?

harmen-xb:
@martin-fleck-at I noticed there was an error in de Windows build output about some missing node module files.

I checked on my laptop and U had the same issue in Windows (which I usually don't use since I usually work in a dev container).

Anyway, because of the issue error message I thought there was something off with with prepare, built and test command for the e2e-test package. So I aligned it more with how it was setup within Theia and that seems to have solved it.

It doesn't actually execute playwright in Windows, but it never did before anyway.

I did have some issues with the GitHub workflow executing successfullym. It failed on some very unrelated things. After 3 executions it now does succeed. So a bit strange, but it looks like it's ok now.

martin-fleck-at:
@harmen-xb Great, thank you very much for looking into that! If the Windows CI is happy, so am I ;-) Please feel free to have a look at the PR once you some time.

harmen-xb:
@martin-fleck-at 

I tested everything after the update and everything seems very responsive and fast. Startup is also noticable faster.

Some issues I found:
- When making a relationship using a system diagram the ID of the edge becomes the same as the ID of the relationship itself. This lways leads to a validation error. I think the ID of the relationship should be suffixed with Edge or something.
- In the mapping diagram when adding a source object the tool should be blocked when hovering over existing nodes. You already implemented this in the system diagram for new entities (you see a forbidden sign when hovering over existing entities).
- In the  mapping diagram when creating a source object the dropdown with entities is openened in the left top corner. We think it's better to open it at the position of the mouse cursor. I know the position of the node will dynamically be defined, but for the user it makes more sense if the dropdown is at the mouse cursor location.

martin-fleck-at:
@harmen-xb Great, thank you for your feedback! I merged the current master into the PR and fixed all the issues that you mentioned if you want to have another look.

harmen-xb:
@martin-fleck-at 

Thanks for the update. The things I noticed last time are indeed all working correct now.

I do see something I didn't notice the previous time (I think it happened after the merge from main).
When I now open the CRM source system diagram I don't see the attribute of the entity.
When I then resize the node I do see them, but the customProperties which are defined on the node are now removed. So I guess something goes wrong there when updating the diagram from the diagram editor versus the text editor.

martin-fleck-at:
@harmen-xb Good catch! I'll have a look at this later today and hopefully it is just a minor issue that happened during the merge.

martin-fleck-at:
@harmen-xb There was indeed a problem with the serialization as we already use 'name' and 'value' in other contexts. And the document for the 'Address' node was not correctly identified. I pushed some commits that should fix those problems. Thank you for catching that!

harmen-xb:
@martin-fleck-at 

The scenario I described works good now. I did find another scenario which isn't working.

If you change an attribute on which customProperties are defined via de property widget, the customProperties are also removed.

Take for example the ExampleCRM.Address entity. If you change the datatype or identifier property of the CountryCode attribute, the customProperties are gone.

martin-fleck-at:
@harmen-xb Another great catch, thank you! Indeed, we manually selected the fields to update which did not include the newly created `customProperties`. I changed it in a way that we hopefully no longer have to manually specify the fields. I also notices that some of our examples were using datatype "Text" which was not part of the form dropdown. I added it so that we get a nicer UX.

extensions/crossmodel-lang/src/language-server/cross-model-serializer.ts
```diff
@@ -214,4 +226,14 @@ export class CrossModelSerializer implements Serializer<CrossModelRoot> {
       const right = obj.expression.right.$type === StringLiteral ? quote(obj.expression.right.value) : obj.expression.right.value;
       return [left, obj.expression.op, right].join(' ');
    }
+
+   private isCustomProperty(obj: any): obj is { name: string; value?: string } {
```
harmen-xb, line 230: @martin-fleck-at 

Can't you check the type of AST node here?

This code looks a bit tricky to me.

harmen-xb (commented):
@martin-fleck-at 

I reviewed the code, all looks good to me. There is just one piece of code that worries me a bit. I put a comment on it.

Keen on hearing your thoughts on this.

extensions/crossmodel-lang/src/language-server/cross-model-serializer.ts
martin-fleck-at: I can look into it but it would mean refactoring the serialization a bit. Currently, we strip out the all properties that are not to be serialized BEFORE we go through the properties so at this point in time we no longer have that information. However, I'm sure that can be improved.

harmen-xb: Ok, let's leave it for now. Maybe this wil change anyway when we upgrade Langium since it now has indent support?

harmen-xb (approved):

extensions/crossmodel-lang/src/language-server/cross-model-serializer.ts
martin-fleck-at: As far as I can tell, we only get the support in grammar and parsing but not in serialization since there is no generic model serialization yet. But I'll know for sure once I look deeper into it.

PR #61, opened 2024-07-28
Feature/composite editor

tortmayr:
Add composite editor and upgrade framework

Add composite editor:
- Introduce new composite-editor package
- Register composite-editor as default open handler
- Composite editor shows different subeditors for the same file in a tabpanel and is enables for 
   - Entities (Form/Code)
   - Relationships (Form/Code)
   - Systemdiagram (Diagarm/Code)
   - Mapping diagram (Diagram/Code)

Upgrade Frameworks:
- Update to GLSP 2.2.1 - the latest GLSP version

Requires https://github.com/CrossBreezeNL/crossmodel/pull/60


tortmayr:
# Known issues:

With the composite editor both the graphical/form and text editor are open at once.
Potentially also the properties view is currently open.
This causes sync issues on change and save

- Changes in the text editor are executed after every keyboard stroke -> this means the graphical editor could receive updates
  for an incomplete/faulty model
- Updates are not debounced -> text editing causes a lot of intermediate updates in the graphical editor
- The graphical diagram has its own commandstack that can get out of sync with a mix of graphical and texting
- Editing something essential in the text editor like the entity property/id of an entry can cause a crash of the graphical server
- For the form editor changes and undo/redo are only applied locally and synced with the model service on save

# Side issue that should be fixed as part of the stabilization efforts
- The CSS of the form editor leaks into the main application and overrides things like font size and icon spacing
- -> Whenever a form editor is open the styling of the Theia application e.g. explorer view changes. Once closed its resets to the original values

tortmayr:
Opened on accident, should be a draft PR in the local ES fork for now

PR #62, opened 2024-08-08
Support custom properties in textual representation

martin-fleck-at:
- Add optional custom properties to most elements in the grammar


harmen-xb (approved):
Changes look good to me.

harmen-xb:
@martin-fleck-at 

The custom properties implementation works exactly as intentent, nice!

I didn't merge yet, as there is some macos built error. But I guess that might also be solved with the changes from the other PR. So I leave it to you whether this can be merged after that one.

PR #63, opened 2024-08-22
Add composite editor and upgrade GLSP to latest version

martin-fleck-at:
Add composite editor containing all other editors
- Introduce new composite-editor package
- Register composite-editor as default open handler
- Composite editor shows different sub-editor depending on file
-- Entities: Form and Code
-- Relationships: Form and Code
-- System Diagrams: Diagram and Code
-- Mapping Diagrams: Diagram and Code

Additional:
- Debounce graphical updates when receiving updates in GLSP server
- Ensure MUI global styles do not destroy Theia styling
- Update to latest GLSP version (v2.2.1)


martin-fleck-at:
@harmen-xb Not sure why the MacOS build keeps failing but I'll keep re-triggering in case it is outside of our control for now and will check up on it tomorrow again.

martin-fleck-at:
@harmen-xb Hopefully fixed it by providing the secret and increasing the rate limit from 60/hour to 1000/hour supposedly.

harmen-xb:
@martin-fleck-at 

The changes look pretty good, the tabbing perspectives is nice.

I do notice some strange behaviour when working in the code editor. When I for example open the 'CalcAge' entity and then edit the file which would make the file invalid according to the grammer. Untill savind it's fine, but when you press Save it removes all parts in the file which are invalid. Why don't we leave the file in a invalid state for the user to be fixed?

martin-fleck-at:
> @martin-fleck-at
> 
> The changes look pretty good, the tabbing perspectives is nice.
> 
> I do notice some strange behaviour when working in the code editor. When I for example open the 'CalcAge' entity and then edit the file which would make the file invalid according to the grammer. Untill savind it's fine, but when you press Save it removes all parts in the file which are invalid. Why don't we leave the file in a invalid state for the user to be fixed?
> 
> Example of a change, remove the double quotes around the name of an attribute (which is incorrect but fixable by the end-user). When you the press Save the whole name part is gone.
> 
> Or when you remove the double quotes around the name of the entity, the name element is gone, nut also everything else in the files. So it looks like the serializer stops on an error but still returns the fill as far as it could serialize it. But in this case it should not serialize it again, just leave the file as is and leave the errors in.

Interesting, I'll have a deeper look at this tomorrow but my best guess is that the parser actually only creates parts of the internal model (as far as it can) and that is what we in the end serialize since we always serialize the in-memory representation.

harmen-xb:
> > @martin-fleck-at
> > The changes look pretty good, the tabbing perspectives is nice.
> > I do notice some strange behaviour when working in the code editor. When I for example open the 'CalcAge' entity and then edit the file which would make the file invalid according to the grammer. Untill savind it's fine, but when you press Save it removes all parts in the file which are invalid. Why don't we leave the file in a invalid state for the user to be fixed?
> > Example of a change, remove the double quotes around the name of an attribute (which is incorrect but fixable by the end-user). When you the press Save the whole name part is gone.
> > Or when you remove the double quotes around the name of the entity, the name element is gone, nut also everything else in the files. So it looks like the serializer stops on an error but still returns the fill as far as it could serialize it. But in this case it should not serialize it again, just leave the file as is and leave the errors in.
> 
> Interesting, I'll have a deeper look at this tomorrow but my best guess is that the parser actually only creates parts of the internal model (as far as it can) and that is what we in the end serialize since we always serialize the in-memory representation.

I get that, but I think if the parser fails it should not serialize it. If the edit comes from the code editor it shouldn't serialize it anyway I think.

I guess this is also related to if the file is invalid, the other perspectives should be disabled untill the file is fixed I think. This definetly a new thing, so not for now.

martin-fleck-at:
@harmen-xb I fixed the issue you mentioned and also introduced a readonly mode for our widgets to notify the user.

harmen-xb:
@martin-fleck-at 

The code editor now works as expected, I can type anything I want (also grammitcally incorrect) and save it. And the form/diagram nicely shows the error about a broken model.

2 things:
- The message sais the "Model" has errors. I am not sure whether end-users understand this phrase. Maybe it's better to say "The file conains one or more errors. Please fix the error(s) using the 'Code Editor'. This perspective will be read-only untill the errors are resolved."
- When I now use the form perspective to change an entity (for example). And I type some text and do "Undo" then it only works once. I thought I could do undo longer to go back in the changes I made.

martin-fleck-at:
@harmen-xb I updated the message and fixed the handling in the form editor. Indeed, that was a problem that was already present before this change that in the form editor you could only undo once. However, now it should be possible to undo at least all the changes done by the form editor itself.

harmen-xb:
@martin-fleck-at The changes look good.

I do think you use the wrong log level for writing the ModelReducer en form-editor outputs. There is a bit to much info logging now I think :).

martin-fleck-at:
@harmen-xb I reduced the log level to debug. If you agree with that, feel free to approve and merge the PR

extensions/crossmodel-lang/src/model-server/model-service.ts
```diff
@@ -92,21 +94,24 @@ export class ModelService {
...
-    * If the document was not already open for modification, it will be opened automatically.
-    *
-    * @param uri document URI
-    * @param guard guard function to ensure a certain type of semantic model
-    */
-   request<T extends AstNode>(uri: string, guard: (item: unknown) => item is T): Promise<T | undefined>;
-   async request<T extends AstNode>(uri: string, guard?: (item: unknown) => item is T): Promise<AstNode | T | undefined> {
+   async request(uri: string, state = DocumentState.Validated): Promise<AstCrossModelDocument | undefined> {
       const document = this.documents.getOrCreateDocument(URI.parse(uri));
+      // should be replaced once we upgrade to newer Langium versions so we can use documentBuilder.waitUntil or the waitUntilPhase function
```
harmen-xb, line 101: Should we prefix this comment with todo? So we can find it later in a todo list or something?

Or what is convention in JavaScript world for this? 🤔

harmen-xb (approved):
One small comment on a comment 😄

extensions/crossmodel-lang/src/model-server/model-service.ts
martin-fleck-at: I'll tackle this with the next PR if that is fine with you since that already is the Langium upgrade ;-)

martin-fleck-at: In general pre-fixing with `TODO:` is better of course but it is already resolved in https://github.com/CrossBreezeNL/crossmodel/pull/64/files#diff-bfc5394600f3b18d3c7ec47fd842312acb8f3072dbf5a142c602b0d3e87ac792L101

PR #64, opened 2024-09-04
Upgrade to latest Langium version

martin-fleck-at:
- Adapt to latest version of Langium
- Re-use indentation-based parsing provided by Langium
- Rework serialization
- Apply and deactivate async threaded parsing as it crashes on error


martin-fleck-at:
This is partially based on https://github.com/CrossBreezeNL/crossmodel/pull/63 and needs to be rebased once the other PR is merged. This is why it is still in draft stage.

harmen-xb (approved):
Code changes look nice, a lot less :).

Also the 2 issues we found in the previous PR before you pushed the bugfix are indeed solved.

Issue #48, opened 2024-02-28
Adding attribute via entity form results in incorrect entity file


harmen-xb:
When adding a new attribute using the entity form/property view the new attribute is created, but doesn't have an Id, so there is an error that the Id is missing.

**Steps to reproduce**
1. Open the form editor on an entity
1. Add an attribute to the entity on the 'Attributes' tab

**Expected behaviour**
The Id should be in the form for now, maybe automatically populated based on the name?

harmen-xb:
We would like to solve this issue genericly for all objects.

Ideally all objects have a Name and an Id. When a new object is created the Id is initially derived from the Name with:
Id = <Name-without-withspace-and-specials> (replace whitespace with underscore and special charachters with closes ANSI character. For example replace á with a.)

When the name is changed, the Id is NOT changed. This way anyrhing referencing the object is not broken.

In the future we can add logic to refactor the Id's if modellers want to.

harmen-xb:
Retested this on latest main and the issue is a bit different now.

When creating new attributes on an Entity the first Id is always 'Attribute'. The next Id is 'Attribute1', next is 'Attribute12', next is 'Attribute 123'.

This is a bit different then expected. It should increase the counter. So first is Attribute, then Attribute2, then Attribute3, then Attribute4.

Also the word 'Attribute' is now fixed. No matter what the user enters in the form for adding attributes the Id always starts with 'Attribute'. This should be different, please see the previous comment for the expected behaviour:

> We would like to solve this issue genericly for all objects.
> 
> Ideally all objects have a Name and an Id. When a new object is created the Id is initially derived from the Name with: Id = (replace whitespace with underscore and special charachters with closest ANSI character. For example replace á with a.)
> 
> When the name is changed, the Id is NOT changed. This way anyrhing referencing the object is not broken.
> 
> In the future we can add logic to refactor the Id's if modellers want to.

harmen-xb:
@martin-fleck-at 

Changes look good, the behaviour on attribute is indeed as expected.

Two things I notice:
- When creating an entity via the diagram, I cannot type the entity name before the actual entity is created (like I can via explorer, or like when adding attribute via the form). This means the ID of the Entity will be NewEntity, NewEntity1, etc.
- When creating an entity with the ID NewEntity, adding some attributes and saving. And then removing the Entity, save. And then add a new Entity 'NewEntity' the deleted Entity is kinda re-used somehow, bevause the NewEntity has the attributes from the removed entity.

PR #69, opened 2024-10-07
Fix auto-naming counter and align id with name when creating new entity

martin-fleck-at:
- Add tests for utility functions

Fixes https://github.com/CrossBreezeNL/crossmodel/issues/48

### DIFF
--- a/extensions/crossmodel-lang/src/glsp-server/common/cross-model-command.ts
+++ b/extensions/crossmodel-lang/src/glsp-server/common/cross-model-command.ts
@@ -12,7 +12,9 @@ import { CrossModelSourceModel, CrossModelState } from './cross-model-state.js';
 export class CrossModelCommand extends JsonRecordingCommand<CrossModelSourceModel> {
    constructor(
       protected state: CrossModelState,
-      protected runnable: () => MaybePromise<void>
+      protected runnable: () => MaybePromise<void>,
+      protected undoAction?: () => MaybePromise<void>,
+      protected redoAction?: () => MaybePromise<void>
    ) {
       super(state, runnable);
    }
@@ -33,6 +35,7 @@ export class CrossModelCommand extends JsonRecordingCommand<CrossModelSourceMode
    override async undo(): Promise<void> {
       if (this.undoPatch) {
          const result = this.applyPatch(await this.getJsonObject(), this.undoPatch);
+         await this.undoAction?.();
          await this.postChange?.(result.newDocument);
       }
    }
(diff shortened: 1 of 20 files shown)


harmen-xb:
@martin-fleck-at 

Changes look good, the behaviour on attribute is indeed as expected.

Two things I notice:
- When creating an entity via the diagram, I cannot type the entity name before the actual entity is created (like I can via explorer, or like when adding attribute via the form). This means the ID of the Entity will be NewEntity, NewEntity1, etc.
- When creating an entity with the ID NewEntity, adding some attributes and saving. And then removing the Entity, save. And then add a new Entity 'NewEntity' the deleted Entity is kinda re-used somehow, bevause the NewEntity has the attributes from the removed entity.

martin-fleck-at:
@harmen-xb Thank you for the review! The first thing I can re-produce. I'll see what we can do about that in the short-term to make that nicer. However, the second point, i cannot really reproduce. Does that happen in the diagram or what is the exact workflow here?

harmen-xb:
@martin-fleck-at 

To reproduce:
1. Create an entity via a system diagram without changig the name.
1. Add some attributes to the entity and save.
1. Remove the new entity from the diagram and object explorer.
1. Add a new Entity via the diagram without specifying a name

Obervation: In the entity created in the last step attributes are already there as if the deleted entity object is re-used.

martin-fleck-at:
@harmen-xb I had a deeper look into this and managed to fix both problems.

harmen-xb:
@martin-fleck-at This indeed works better :)

Is it possible to write the new entity file after the name is typed in the diagram? Cause now when you add a new entity to a diagram the 'NewEntity' file is already created (before typing the real name of the entity). Would be nice if file name would be the same as the ID.

martin-fleck-at:
@harmen-xb I think I could look into that but it would be like developing a complete new tool so it might be some effort. Would it be an idea to simply rename the file according to the new name, similar to what Java usually proposes?

harmen-xb:
> @harmen-xb I think I could look into that but it would be like developing a complete new tool so it might be some effort. Would it be an idea to simply rename the file according to the new name, similar to what Java usually proposes?

@martin-fleck-at Yes, if you could rename the file if the `Id `changes that would be great! This would in the end be a generic behaviour for all files to rename the file if the Id changes. So we could also pick this up in a separate branch.

martin-fleck-at:
@harmen-xb Great, should we create a ticket for this and then merge this PR?

harmen-xb (approved):
Code changs look good.

harmen-xb:
> @harmen-xb Great, should we create a ticket for this and then merge this PR?

Yes, I reviewed code and approved it.

Issue #66, opened 2024-09-20
System diagram show entity lists double


harmen-xb:
When using the 'Show Entity' tool on a system diagram it lists the entities twice, once name only and once fully qualified.

![image](https://github.com/user-attachments/assets/d4b4d47a-e3b8-47b7-86b2-42f53c6dccb9)

Preferred way is to have the same behaviour as textual editor.

PR #70, opened 2024-10-07
Remove duplicate entries for auto-completion suggestions and sort

martin-fleck-at:
- Prefer package local names over global names
- Sort local names before global name in textual and diagram editor

Fixes https://github.com/CrossBreezeNL/crossmodel/issues/66

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -115,12 +115,9 @@
       "name": "Debug All CJS Jest Tests",
       "type": "node",
       "request": "launch",
-      "runtimeArgs": [
-        "--inspect-brk",
-        "${workspaceRoot}/node_modules/jest/bin/jest.js",
-        "--runInBand",
-        "--config=${workspaceRoot}/configs/jest.config.js"
-      ],
+      "cwd": "${workspaceRoot}",
+      "runtimeArgs": ["--inspect-brk", "node_modules/jest/bin/jest.js"],
+      "args": ["--runInBand", "--config=configs/jest.config.js"],
       "console": "integratedTerminal",
       "internalConsoleOptions": "neverOpen"
     },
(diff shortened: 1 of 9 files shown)


martin-fleck-at:
@harmen-xb I'm facing an issue with Windows on this one. Maybe you could help me out here since I don't have a Windows machine to try. It must have something to do with the paths not properly recognizing that one URI is not a child of another.

harmen-xb:
@martin-fleck-at I finally got the debugging to work (i had the same issue on linux when running a single test). It had to do with the 'outFiles' being set to an empty array apparantly.

I ran the unit tests of the uri-util class and I noticed a relavant difference in the posifxPath.relative function.

See the relative result in linux:
image

And on Windows:
image

See how in Windows the result is prefixed with ,,\, i.s.o. only the child part. I have the idea the problem is in this part.

martin-fleck-at:
> [UPDATE] I checked the uri-utils, what would happen if we only use nodePath and not the posix variant if it's available and then all tests on the uri-utils succeed on linux and on windows. Also I see the relative command working. I think this is because the unit tests only use posix paths, so there is no file:\C:\some\path as an example. It does however not solve the label issue (I can't find where the label is constructed in the code).

That sounds good! So if I understand you correctly, not using the posix-path yields better result and for the test URIs we should use Windows-style paths on Windows and Unix-style paths on Unix and then it should work, is that correct?

harmen-xb:
> That sounds good! So if I understand you correctly, not using the posix-path yields better result and for the test URIs we should use Windows-style paths on Windows and Unix-style paths on Unix and then it should work, is that correct? I do not really remember anymore why we had to do the posix-specific behavior tbh.

The uri-utils tests do succeed on windows when you remove the usage of posix. But as I said, I think this is mainly because there is no test with a real Windows path starting with a drive. It might fail at that point, didn't have time to test this yet.

Maybe we can have a look at the code where the label of an object is constructed and debug it there. Cause on Windows it's stil not prefixed with the system node (like in Linux)..

martin-fleck-at:
@harmen-xb The latest update made the CI happy so I think we can merge this, what do you think? Feel free to approve if you agree ;-)

harmen-xb (approved):
Changes look good!

Issue #65, opened 2024-09-20
Error message on diagram and forms


harmen-xb:
An error message is presented on a form or diagram when the underlaying file is broken.

- For diagrams now only a warning icon is shown on which you need to hover to see the error.
- In forms the error message is presented very small.

It would be better if the error message is better readable and always shown (no need to hover).

PR #71, opened 2024-10-17
Centralize error message and make it display it more prominently

martin-fleck-at:
- Create a central EditorError component and use in graphical and form
- Remove message coming from GLSP server and replace with UI extension
- Ensure tool palette minimize button is shown depending on state

Fixes https://github.com/CrossBreezeNL/crossmodel/issues/65


harmen-xb (approved):
Nice change including the playwright tests!

PR #74, opened 2024-11-25
Refactored build to separate browser and electron and added docker build for main

harmen-xb:
- Refactored main build script to have browser and electron build (incl. symlink) to have minimize build for build and docker.
- Updated all versions to 0.0.0.
- Created Dockerfiles for ubuntu and alpine and added/updated build scripts
- Setup docker build in main pipeline, using the alpine Dockerfile.


harmen-xb:
@martin-fleck-at hereby the PR for Docker container. I removed the rimraf in node_modules as discussed.

I also made two build options for browser and electron to only build what is necessary.

package.json
```diff
@@ -10,14 +10,20 @@
   ],
   "scripts": {
     "build": "lerna run build",
-    "clean": "lerna run clean && rimraf node_modules",
+    "build:browser": "yarn build:packages && yarn build:extensions && yarn symlink:browser && yarn theia:browser build",
```
martin-fleck-at, line 13: If we are already splitting the build into browser and electron, maybe we should do the same for watching, what do you think?
```
    "watch:browser": "lerna run --parallel --ignore=\"crossmodel-app\" watch",
    "watch:electron": "lerna run --parallel --ignore=\"crossmodel-browser-app\" watch"
```

Watching is very fast in general so it may not be necessary but adding those two scripts also doesn't hurt I think.

Dockerfile.ubuntu
```diff
@@ -0,0 +1,91 @@
+# Stage 1: Builder stage
```
martin-fleck-at, line 1: Could you explain why we have two Dockerfiles here? When do we need the `.ubuntu` one? Might be good to add to the a README or add a comment somewhere in the file as it might be non-obvious.

.github/workflows/cicd-main.yml
```diff
@@ -60,3 +65,49 @@ jobs:
       - name: Deploy to GitHub Pages
         id: deployment
         uses: actions/deploy-pages@v4
+
+  # Create Docker image with CrossModel web application.
+  docker:
```
martin-fleck-at, line 70: That is a bit hard to test as we only trigger it on pushed to the main branch so I'll let the CI decide the final fate once we merge but overall the code looks good to me ;-)

martin-fleck-at (commented):
The change looks good to me overall and I like the unification of the version numbers as well as the splitting between browser and electron app. I built and ran the Docker image locally and it seemed to work as expected. I just have a few minor comments which maybe could be addressed.

.github/workflows/cicd-main.yml
harmen-xb: Yes, I tested it in feature and then moved to main. So I pretty confident it works :).

Dockerfile.ubuntu
harmen-xb: I added explanation about this in the README.

package.json
harmen-xb: Good point, added the specific watch scripts and updated the README.

martin-fleck-at (approved):
All my concerns have been addressed, thank you for the quick turnaround, Harmen! I'll ask Johannes tomorrow to just have a very quick look at it in case he spots something odd but from my side we can already merge this and fix any unexpected issues in a follow-up.

Dockerfile.ubuntu
```diff
@@ -0,0 +1,91 @@
...
+FROM node:20-bookworm AS build-stage
```
martin-fleck-at, line 2: Just realized that this is not actually Ubuntu but Debian, at least according to https://hub.docker.com/_/node:
> Some of these tags may have names like bookworm or bullseye in them. These are the suite code names for releases of [Debian⁠](https://wiki.debian.org/DebianReleases) and indicate which release the image is based on.

This is not a problem, just makes the file naming a bit confusing. Sorry, I missed that yesterday!

PR #77, opened 2024-12-03
Add, edit and delete entity inheritances in the text representation

martin-fleck-at:
- Extend grammar with 'inherits' keyword and super entities
- Ensure proper serialization and cycle detection validation
- Add test cases

**Please note** that this is based on https://github.com/CrossBreezeNL/crossmodel/pull/76 as they both touch the validator. So https://github.com/CrossBreezeNL/crossmodel/pull/76 should be merged first.


harmen-xb (approved):
@martin-fleck-at

Code changes look good.

I added the example inheritance files from the branch we prepared with the examples. I think this is a nice example to use to implement the inheritance edges for the next story.

PR #79, opened 2024-12-19
Setup clean default layout for application

martin-fleck-at:
layout


martin-fleck-at:
@harmen-xb Pushed an update to add the source control view to the default layout, add the Git extension by default and support extension installation from Open VSX.

harmen-xb (approved):
Code changes look good.

PR #90, opened 2025-01-08
Add entity inheritances in the diagram representation

tortmayr:
- Render entity inheritances as entity edges
- Extend edge creation tool to enable creation of new inheritances
- Add inheritance creation tool to tool box
- Add operation/command handlers on server side for inheritance creation & deletion

Also:
- Update tsconfig of vscode extension to fix ts errors related to inversify property injection
- Fix wrong entity initialization in create operation handler

### DIFF
--- a/e2e-tests/src/page-objects/cm-app.ts
+++ b/e2e-tests/src/page-objects/cm-app.ts
@@ -7,6 +7,7 @@ import { TheiaEditor, TheiaNotificationIndicator, TheiaNotificationOverlay, Thei
 import { CMCompositeEditor, IntegratedEditorType } from './cm-composite-editor';
 import { CMExplorerView } from './cm-explorer-view';
 import { CMTheiaIntegration } from './cm-theia-integration';
+import path = require('path');
 
 export interface CMAppArgs extends Omit<IntegrationArgs, 'page'> {
    workspaceUrl?: string;
@@ -18,7 +19,7 @@ export class CMApp extends TheiaGLSPApp {
          { browser: args.browser, page: {} as any, playwright: args.playwright },
          {
             type: 'Theia',
-            workspace: args.workspaceUrl ?? 'src/resources/sample-workspace',
+            workspace: args.workspaceUrl ?? path.join(__dirname, '../resources/sample-workspace'),
             widgetId: '',
             url: args.baseUrl ?? 'http://localhost:3000'
          }
--- a/examples/mapping-example/ExampleDWH/diagrams/ProductInheritance.system-diagram.cm
+++ b/examples/mapping-example/ExampleDWH/diagrams/ProductInheritance.system-diagram.cm
@@ -7,13 +7,13 @@ systemDiagram:
         x: 242
         y: 33
         width: 154.649658203125
-        height: 116
-      - id: ExampleDWH_NewEntityNode
+        height: 116.11013984680176
+      - id: ExampleDWH_DigitalProductNode
         entity: ExampleDWH.DigitalProduct
-        x: 187
+        x: 132
         y: 275
-        width: 154.22557544708252
-        height: 73.93824672698975
+        width: 154.3004560470581
+        height: 96.0881118774414
       - id: PhysicalProductNode
         entity: PhysicalProduct
         x: 429
@@ -22,13 +22,26 @@ systemDiagram:
         height: 136
       - id: PerishableProductNode
         entity: PerishableProduct
-        x: 429
+        x: 583
         y: 495
-        width: 155.22567415237427
-        height: 73.93824672698975
+        width: 183.74362182617188
+        height: 76
       - id: SomeProductNode
         entity: ExampleOtherModel.SomeProduct
-        x: 671
+        x: 682
         y: 352
         width: 207.14583730697632
-        height: 55.33333396911621
+        height: 56.0440559387207
+    edges:
+      - id: ExampleDWH_DigitalProductNodeInheritanceEdge
+        baseNode: ExampleDWH_DigitalProductNode
+        superNode: ExampleDWH_ProductNode
+      - id: PhysicalProductNodeInheritanceEdge
+        baseNode: PhysicalProductNode
+        superNode: ExampleDWH_ProductNode
+      - id: PerishableProductNodeInheritanceEdge
+        baseNode: PerishableProductNode
+        superNode: PhysicalProductNode
+      - id: PerishableProductNodeInheritanceEdge1
+        baseNode: PerishableProductNode
+        superNode: SomeProductNode
--- a/extensions/crossmodel-lang/src/glsp-server/system-diagram/edge-checker/edge-checker.ts
+++ b/extensions/crossmodel-lang/src/glsp-server/system-diagram/edge-checker/edge-checker.ts
@@ -1,0 +1,66 @@
+/********************************************************************************
+ * Copyright (c) 2025 CrossBreeze.
+ ********************************************************************************/
+import { INHERITANCE_EDGE_TYPE } from '@crossbreeze/protocol';
+import { EdgeCreationChecker, GModelElement, ModelState, getOrThrow } from '@eclipse-glsp/server';
+import { inject, injectable } from 'inversify';
+import { Entity, EntityNode, isInheritanceEdge } from '../../../language-server/generated/ast.js';
+import { SystemModelState } from '../model/system-model-state.js';
+
+@injectable()
+export class SystemEdgeCreationChecker implements EdgeCreationChecker {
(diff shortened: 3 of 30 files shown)


harmen-xb:
@martin-fleck-at @tortmayr 

Thanks for the update!

I updated the inheritance system diagram to include the existing inheritances (using the 'Create inheritance' tool and that worked good 👍 ).

I tested the workings and I have some findings:
- When I create an inheritance in the system diagram between the same two nodes twice, it results in 2 inheritance edges representing the same inheritance. This should not be possible, inheritance edge can only exist once between 2 nodes.
- When I create an inheritance in the opposite direction of an existing inheritance, I am allowed to do it in the system diagram. So in the example there is already an inheritance from DigitalProduct to Product. When I then drag from DigitalProduct to Product it actually creates the edge and updates the Product entity. I think we should block this in the diagram by showing this is not possible.
- When you open the ProductInheritance system diagram in the ExampleDWH model (in the mapping-example workspace) and create an inheritance from SomeProduct to Product it doesn't seem to do anything. The other way around works. I first didn't understand why, but now I see SomeProduct resides in another model. So it tries to add an inheritance to SomeProduct refernecing a model which is not in the dependencies list of that model yet. In this case ExampleDWH should be added to the list of dependencies of ExampleOtherModel. This is generic behaviour which also goed wrong when for example creating a diagram in the ExampleOtherModel model and dragging the ExampleDWH.Product entity on the new diagram. This leads to an unresolved node, because the dependency is missing.

martin-fleck-at:
> * When I create an inheritance in the opposite direction of an existing inheritance, I am allowed to do it in the system diagram. So in the example there is already an inheritance from DigitalProduct to Product. When I then drag from Product to DigitalProduct it actually creates the edge and updates the Product entity. I think we should block this in the diagram by showing this is not possible.

I think we discussed this at some point and argued that we do not want to restrict the user in advance but much rather indicate an error through validation. In this example, the user would create an intermediate erroneous state that they need to fix afterwards. If this were not be possible, the user would need to delete the other inheritance first before creating the new one. What do you think? If we do want to restrict the user, we need to figure out a good way to indicate to the user _why_ something is not allowed because simply a "forbidden" cursor might not be enough information, especially if cycles can be very complex or not all nodes in the cycle are present in the diagram.

harmen-xb:
> > * When I create an inheritance in the opposite direction of an existing inheritance, I am allowed to do it in the system diagram. So in the example there is already an inheritance from DigitalProduct to Product. When I then drag from Product to DigitalProduct it actually creates the edge and updates the Product entity. I think we should block this in the diagram by showing this is not possible.
> 
> I think we discussed this at some point and argued that we do not want to restrict the user in advance but much rather indicate an error through validation. In this example, the user would create an intermediate erroneous state that they need to fix afterwards. If this were not be possible, the user would need to delete the other inheritance first before creating the new one. What do you think? If we do want to restrict the user, we need to figure out a good way to indicate to the user _why_ something is not allowed because simply a "forbidden" cursor might not be enough information, especially if cycles can be very complex or not all nodes in the cycle are present in the diagram.

Yes I remember. At that point I was thinking about indirect cycles, so via another node. But this is a direct cycle. Because the edges are overlapping you can't select 1 and remove it.

So either we need to block it or the routing algorithm should change a bit so the lines are not overlapping.

Maybe the routing algorithm change would be best indeed.

tortmayr:
@harmen-xb Thanks for the detailed review. I have updated the PR as follows:

>When I create an inheritance in the system diagram between the same two nodes twice, it results in 2 inheritance edges representing the same inheritance. This should not be possible, inheritance edge can only exist once between 2 nodes.

This is now fixed. A check is in place that prevents the creation of an additional inheritance edge if there is already an edge between the two nodes representing the same inheritance.

>Yes I remember. At that point I was thinking about indirect cycles, so via another node. But this is a direct cycle. Because the edges are overlapping you can't select 1 and remove it.
>
>So either we need to block it or the routing algorithm should change a bit so the lines are not overlapping.
>
>Maybe the routing algorithm change would be best indeed.

Unfortunately a routing algorithm change alone would not be enough. The diagram becomes readonly if there is a validation error. So even if we fix the routing so that the edges don't overlap the error cannot be resolved in the diagram (readonly) and could only be fixed in the textual representation.
Therefore, I opted for blocking the operation if it would result in a direct inheritance cycle. In addition, a warning message is displayed to the user.

 >When you open the ProductInheritance system diagram in the ExampleDWH model (in the mapping-example workspace) and create an inheritance from SomeProduct to Product it doesn't seem to do anything. The other way around works. I first didn't understand why, but now I see SomeProduct resides in another model. So it tries to add an inheritance to SomeProduct referencing a model which is not in the dependencies list of that model yet. In this case ExampleDWH should be added to the list of dependencies of ExampleOtherModel. This is generic behaviour which also goes wrong when for example creating a diagram in the ExampleOtherModel model and dragging the ExampleDWH.Product entity on the new diagram. This leads to an unresolved node, because the dependency is missing.
 
 I would suggest to tackle this issue in a follow-up task. As you already mentioned this more of a generic issue that also affects other aspects of diagram editing (like entities). Identifying a missing dependency and the corresponding model that needs to be added is not a trivial task. We should come up with a clean generic solution that works for all cases.
Another thing that we have to consider here are cyclic dependencies.
For instance, looking at the mapping example:
The `ExampleDWH` model already has a dependency to `ExampleOtherModel`. By adding an inheritance from `SomeProduct` to `Product` we would also need a dependency for `ExampleOtherModel` to `ExampleDWH`.

harmen-xb:
Hi @tortmayr ,

Thanks for the update!

For both not allowing to create an inheritance edge twice and not allowing a direct cyclic inheritance, can we use show it visually in the diagram by showing the stop-sign icon when we or hovering on a target which is not allowed?

I think this is more user friendly, than not showing anything (when creating the inheritance twice) or showing a warning afterwards.

This behaviour should be equal to how it's implemented in the mapping diagram when using the 'Create Mapping' tool. The stop-sign icon disappears when we are on a target where it is allowed to let go.

harmen-xb:
@tortmayr I just had a chat with @martin-fleck-at and the inheritance action where you create an inheritance on a remote entity in the diagram is the only action we have where we need to added the dependencies list.

For now I would rather block the action where the child side of the inheritance is a remote entity. It should block just like with creating the same inheritance twice.

tortmayr:
@harmen-xb I have updated the PR as discussed and added visual validation to the `SystemEdgeCreationTool`.
The following cases are now disallowed
- Create an inheritance edge if there is already an existing edge representing the same inheritance
- Creation of inheritance edges that would introduce a direct cycle
- Creation of inheritance edges where the child side of the inheritance is and unresolved remote entity


https://github.com/user-attachments/assets/4dd0c021-6ee9-4bea-8764-2ec379a4884b



I also discovered a significant performance issue with the `SystemEdgeCreationTool`. Previously during the creation process the tool did continously send feedback update actions and checked if the connection is allowed whenever the mouse was moved. This resulted in a lot of unnecessary overhead as we only need to do these checks and feedback updates when the potential connection candidate changes (i.e. whenever the element under the mouse changes).
This is now fixed but required a refactoring of the `SystemEdgeCreationTool`. Could you please look out for any potential regression issues during testing i.e. make sure that the tool still behaves as expected.

harmen-xb:
@tortmayr 

Thanks for the update! The behavior of the inheritance tool looks good now. I tested a bit and I don't see strange things in the UI. Did you implement the optimization genericly for all edge tools? So also for the create mapping tool in the mapping diagram?

Also, if you create branches could you prefix them with 'feature/' cause the the CI/CD pipeline is automatically ran at every commit. I noticed renaming doesn't work well in GitHub, but a unit test is failing: https://github.com/CrossBreezeNL/crossmodel/actions/runs/13134077562/job/36645443260

tortmayr:
>Thanks for the update! The behavior of the inheritance tool looks good now. I tested a bit and I don't see strange things in the UI. Did you implement the optimization genericly for all edge tools? So also for the create mapping tool in the mapping diagram?

No, I only optimized the system edge creation tool. For the create mapping tool this is not necessary. The difference is that the system edge creation tool is a complete custom tool, while the create mapping tool extends the default GLSP `EdgeCreationTool` (which already has this performance tweaks in place).

>Also, if you create branches could you prefix them with 'feature/' cause the the CI/CD pipeline is automatically ran at every commit. I noticed renaming doesn't work well in GitHub, but a unit test is failing:

I was not aware of the naming convention. Thank you, I will keep it in mind for future PRs.
I pushed an update that should fix the failing unit test and also adds an additional test for the diagram inheritance edges.

harmen-xb (approved):
Code changes look good to me.

Issue #82, opened 2024-12-20
When editing in the entity form i keep getting messages that file has changed


willem-xbreeze:
Created a new entity via the context menu. When typing a description i repeatedly get the message that the entity was modified on the file system.

![image](https://github.com/user-attachments/assets/b77a6c83-4910-46cf-8970-b5a8abca8ddc)

harmen-xb:
@martin-fleck-at 

I send some e-mail about this issue, but apparantly Willem already reported it in december. This seems similar behaviour as I am experiencing right now.

To reproduce:
- Open the example-mapping workspace
- Open any entity using the default (composite editor).
- Go to the 'Code' tab of the composite editor to change the file
- Try to save it, it will revert the change back (most of the time, sometimes it works).

harmen-xb:
@martin-fleck-at @tortmayr I merged the PR with the sync fix, which seemed to work fine in the feature branch. I also updated the pdm-grammar branch to verify whether everything is ok.

But now in main and also in my pdm-grammar feature branch I get issues where it looks like the sync is not happening correctly.
I get errors in the scenarios:
- cm-error-view.spec.ts:14:8 › CrossModel Error Views › Form Editor should show error if model code is broken 
- form/add-edit-delete-entity.spec.ts:54:8 › Add/Edit/Delete entity from explorer › Edit entity name & description using form editor

I can replicate this behavior on my machine, where I can see when switching between the perspectives of the composite editor the change isn't effectuated in these scenarios.

You can see it in the workflow run of the pdm-grammar branch here: https://github.com/CrossBreezeNL/crossmodel/actions/runs/13459383665/attempts/1

You can see the same scenario also fail in main after the PR merge: https://github.com/CrossBreezeNL/crossmodel/actions/runs/13459272361

Apparently it's not always happening, since in the feature branch it was green.

The follow-up scenarios also fail, which I think makes sense because the previous scenario failed.

Can one of you have a look at this?

harmen-xb:
@martin-fleck-at Strange, in the pdm-grammar build I also don't have the "form/add-edit-delete-entity.spec.ts:54:8 › Add/Edit/Delete entity from explorer › Edit entity name & description using form editor" scenario failing anymore. I don't like this type of behaviour.

Let's close it for now.

harmen-xb:
@martin-fleck-at 

I am running the playwright tests on my machine locally and I do see this test failing sometimes. Not all the time.

[...]

PR #96, opened 2025-02-06
Improving syncing between editors on save

martin-fleck-at:
- Allow flag to skip overwrite dialog
- Only save a single editor to save the file only once

Fix https://github.com/CrossBreezeNL/crossmodel/issues/82


tortmayr:
@harmen-xb Yes, I'm currently investigating the playwright test issues

tortmayr:
@harmen-xb I have pushed an update to the PR that fixed the remaining issues. E2E tests are now working as expected.

PR #106, opened 2025-02-28
Improve syncing behavior by ensuring we use Monaco version number

martin-fleck-at:


.github/workflows/build-and-test.yml
```diff
@@ -18,7 +18,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
```
harmen-xb: @martin-fleck-at can you undo this workflow change so all os'es are testen again?

Thanks!

harmen-xb (approved):
Code changes look good, I reverted the temp workflow change.

.github/workflows/build-and-test.yml
martin-fleck-at: I see you already did that, thank you!

martin-fleck-at:
@harmen-xb I see you already finalized it, is that correct? I think all the other code changes, we want to keep.

PR #108, opened 2025-03-05
Updated langium grammar with supertypes 

harmen-xb:
First step to refactoring grammar to support multiple datamodel types. This PR contains adding supertypes to the grammar and some minor property changes.


harmen-xb:
@martin-fleck-at Can you have a look at the failing form error playwright test.

I had this issue on my machine as well. I was hoping it would be fixed with the updates from main, but apparently I broke something in this branch.

extensions/crossmodel-lang/src/glsp-server/mapping-diagram/handler/create-edge-operation-handler.ts
```diff
@@ -19,7 +19,7 @@ import { MappingModelState } from '../model/mapping-model-state.js';
 
 @injectable()
 export class MappingEdgeCreationOperationHandler extends JsonCreateEdgeOperationHandler {
-   @inject(ModelState) protected override modelState!: MappingModelState;
+   @inject(ModelState) protected declare modelState: MappingModelState;
```
colin-grant-work, line 22: This change has been made in a number of files. It doesn't look quite correct to me: `declare` is usually reserved for declaration files or phantom fields, but `modelState`  is definitely supposed to be a real entity here. What is the reason for the change?

colin-grant-work:
@martin-fleck-at, @harmen-xb, I've pushed a commit that updates the form when the diagnostics on a document change, and I believe that should fix the test failure. It looks like previously, the parser was fragile enough that adding an error would change the parse result (e.g. deleting `attributes` from an `Entity`) and the old check for differences in the structure of the document would trigger a rerender, but @harmen-xb's changes have made the parser more robust (it seems) so that we don't get the same error-conditioned parse fails. If you think that that shouldn't be happening and that the document parsing should be breaking harder in the presence of errors, that might be something to look into.

The new check does produce more rerenders, since the diagnostics can differ just because the range changes because you typed one more character as part of your erroneous code. We could be more targeted and compare whether there are new errors, for example, if you believe that's preferable.

harmen-xb:
@colin-grant-work Thanks for the fix.

Can you explain in a bit more detail why the parsing isn't failing anymore? Cause I don't understand why my changes would change the expected behaviour.

The goal of the error in the form view is if someone would edit the file in the code editor and creates an invalid model state (something which can't be parsed) the form view is blocked and shows there is an error in the file.

If the file can be parsed, I think we can render a valid form so we don't need an error message.

Is the diagnostic the result of validations? Cause then we could (in the future) add error messages next to the erroneous field in the form.

packages/core/src/browser/model-widget.tsx
```diff
@@ -123,7 +123,7 @@ export class CrossModelWidget extends ReactWidget implements Saveable {
    }
 
    protected async handleExternalUpdate({ document, reason, sourceClientId }: ModelUpdatedEvent): Promise<void> {
-      if (this.document && !deepEqual(this.document.root, document.root)) {
+      if (this.document && (!deepEqual(this.document.root, document.root) || !deepEqual(this.document.diagnostics, document.diagnostics))) {
```
martin-fleck-at, line 126: Good catch! And I'm always happy if a test case actually finds a problem :D

extensions/crossmodel-lang/src/glsp-server/mapping-diagram/handler/create-edge-operation-handler.ts
martin-fleck-at: I think I introduced this at some point since otherwise we got errors that the field was not initialized unless we added the definite assignment assertion (`!`) of which I'm not the biggest fan. I understand the argument that `declare` is also not the best way. Curiously, building and linting the application now, I don't see the error anymore, even without the `!`, so we could restore those statements.

colin-grant-work:
> Can you explain in a bit more detail why the parsing isn't failing anymore? Cause I don't understand why my changes would change the expected behaviour.

@harmen-xb, I can describe the symptoms; I'm not sure I can say with much confidence why the behavior is different. I was testing with entities in the `mapping-example` like [this one](https://github.com/CrossBreezeNL/crossmodel/blob/main/examples/mapping-example/ExampleDWH/entities/PerishableProduct.entity.cm). On `main`, when no errors were present, the document was parsed fully, and the resulting `document.root` in the form component had two attributes. When an error was introduced by adding a line like `break code` after the ID, the document received by the form component had no attributes - it seemed as though the parse just failed after the error, or maybe the serializer took note of errors and transmitted a skeleton? On this branch, although the error was detected, the document still had all of the features of the document before the error, so `document.root` as seen by the form component was unchanged.

My two best guesses for where differences might be are the parser, which is definitely different with the new grammar, or in serialization, which also changed a fair bit, but I'm not familiar enough with the code to point to a definite cause.

martin-fleck-at (approved):
Changes look good to me, let's get this merged!

martin-fleck-at:
@harmen-xb Do you think we can merge this or would you rather dive deeper into why the parser (most likely) is behaving slightly different now?

harmen-xb:
Let's merge it! :)

PR #109, opened 2025-03-05
New Data Model Action & Model Member Validation

colin-grant-work:
This PR adds two sets of new features, the ability to create a new system via a dialog, and basic handling of system types to disable the creation of objects that aren't compatible with a particular system type.

New system dialog:

image

Command enablement depending on system type:

All available in system of type `logical`:

image

Only system (see below) available in system of type `empty`

image

With error on any existing object that is incompatible with its host system:

image

Open questions:

Currently, I've made it possible to add a new system anywhere to support the concept of a subsystem, but if it would be desirable to have systems only available outside of other systems, I can also implement that logic.

### DIFF
--- a/e2e-tests/src/page-objects/cm-new-model-dialog.ts
+++ b/e2e-tests/src/page-objects/cm-new-model-dialog.ts
@@ -1,0 +1,35 @@
+/********************************************************************************
+ * Copyright (c) 2023 CrossBreeze.
+ ********************************************************************************/
+import { TheiaDialog, USER_KEY_TYPING_DELAY } from '@theia/playwright';
+
+export class CMNewModelInputDialog extends TheiaDialog {
+   protected gridSelector = '.new-model-input-grid';
+
+   async enterName(inputValue: string): Promise<void> {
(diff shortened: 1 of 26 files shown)


harmen-xb:
@colin-grant-work Thanks for your first contribution to CrossModel!

I get the feeling based on the contents of the PR you didn't have the most recent contents of the story on our board. So here a screenshot:



Key thing for me is that de concept is not named 'system' but 'data model'.

I think you could change the 'empty' type to 'physical' or 'conceptual' to add a type which will exist in the future.

For the question on nested systems, this is not something I see we want to support. So please adjust this in the implementation.

Thanks!

harmen-xb (commented):
Thanks for applying the changes so fast!

All changes look good to me.

Could you also add a playwright test for creating a new system?

harmen-xb (approved):
@colin-grant-work 

Code looks good, I made 2 comments for which I pushed a couple of commits. Let me know if the changes I made are ok.

colin-grant-work:
@harmen-xb, thanks for the input and the commits. Looks good to me!

PR #110, opened 2025-03-07
Add label to form editor title

colin-grant-work:
At the moment, there is no label on the tabs for form editors if they are not part of a composite editor; this adds one.

## To Test

1. Open an item that can be opened in a form using the 'Open with...' action.
2. The form should appear, and its tab should have a label.


martin-fleck-at (approved):
LGTM!

PR #113, opened 2025-03-11
Renamed EntityAttribute to LogicalAttribute, updated data types and fixed relationship creation and form

### LINKED ISSUE #111: Mismatch between relationship form and grammar
The grammar for a relationship does not specify a type field:

https://github.com/CrossBreezeNL/crossmodel/blob/52b2fc39afb9440ba772d664c95d24fa1812ad11/extensions/crossmodel-lang/src/language-server/grammar/relationship.langium#L18-L35

The form for a relationship does expect a type field:

https://github.com/CrossBreezeNL/crossmodel/blob/52b2fc39afb9440ba772d664c95d24fa1812ad11/packages/react-model-ui/src/views/form/RelationshipForm.tsx#L46-L51

The two should likely expect the same fields, particularly given that the type in the form has a `*` in its placeholder indicating that it may be a required field.

harmen-xb:
As part of the LDM grammar refactoring I renamed EntityAttribute to LogicalAttribute and updated it's data types (and usage), which already caused quite some changes. I will do other refactoring in a follow-up PR to do it step-by-step.

I noticed not all PlayWright tests where running anymore, that's why I changed the test type of the new system test (https://github.com/CrossBreezeNL/crossmodel/commit/cd279365dfaf29da26a806347cbcd0472b200bfb).

I also removed some old example folders (https://github.com/CrossBreezeNL/crossmodel/commit/4365c5e32f283b60d6949f75dd524d3d03e3b54f).

While I was at it I also fixed #111 to get some feeling with the changes needed in the protocol and forms (https://github.com/CrossBreezeNL/crossmodel/commit/f5333c575b408c12be357d79ade9dceb3dbb3203).


packages/react-model-ui/src/views/form/RelationshipForm.tsx
```diff
@@ -60,6 +53,17 @@ export function RelationshipForm(): React.ReactElement {
                selectOnFocus={true}
             />
 
+            <Select
```
martin-fleck-at, line 56: This is true for all the newly introduced `Select` elements: The rendering seems a bit off since it is not taking the full width and the padding is larger than in the rest of the elements in the form:
image
I'd suggest to either use the `AsyncAutoComplete` element like we do with the parent element or at least add those two properties to the `Select` element to get a more uniform UI: `size='small' fullWidth={true}`. Also there seems to be an issue with the label not being properly shown when the dropdown is open.

examples/language-examples/conceptual/relationships/Address_Customer.relationship.cm
```diff
@@ -2,5 +2,6 @@ relationship:
     id: Address_Customer
     name: "Address - Customer"
     parent: Customer
-    child: ExampleCRM.Address
-    type: "1:1"
\ No newline at end of file
+    parentCardinality: one
```
martin-fleck-at, line 5: I'm adding this comment here since I cannot add it to some of the renamed files but for me, this example is not fully valid as many entities have errors about indentation having parts like:
```
identifier:
      - attributes:
        - CustomerID
```

which is not how we currently represent identifier attributes. Probably follow-up PRs might fix that, just want to make sure this is noted.

martin-fleck-at (changes requested):
Changes look good to me for the most part, thank you very much Harmen! However, I did notice that the new Select-UI does not really fit into the form that well and some of the example moved files have errors. Could you please take a look at this?

examples/language-examples/conceptual/relationships/Address_Customer.relationship.cm
harmen-xb: Hi Martin,

This indeed is not in the grammar yet.

This example shows how the end-state of the grammar (logical, conceptual, relational, etc.) should be. My idea was to change this in a future PR.

packages/react-model-ui/src/views/form/RelationshipForm.tsx
harmen-xb: Thanks for the input, will look into it this afternoon.

examples/language-examples/conceptual/relationships/Address_Customer.relationship.cm
martin-fleck-at: I thought so. Works for me since we primarily use the mapping example for local testing anyway. Thank you for clarifying!

packages/react-model-ui/src/views/form/RelationshipForm.tsx
harmen-xb: I refactored the Select to an Autocomplete (since it's a static list and not something we fetch from an API for now).

This solved the look-and-feel and label problem.

We can make it nicer in the future when we invest time in generating the forms.

martin-fleck-at (approved):
LGTM!

PR #114, opened 2025-03-13
Renamed Entity to LogicalEntity and added identifiers property

harmen-xb:
- Refactored Entity to LogicalEntity in grammar, protocol, GLSP and front-end where necessary).
- Added identifiers property to LogicalEntity, where multiple identifiers can be modelled including validation.

_Note: there is still an identifier property on LogicalAttribute, which we need to remove from the grammar and refactor (in the future) to be a calculated property based on the identifiers property on LogicalEntity._


martin-fleck-at (approved):
LGTM!

Issue #81, opened 2024-12-20
When adding a mapping graphical perspective is in error state


willem-xbreeze:
When creating a new mapping via the contextmenu, the graphical perspective is in error state and i have to first create a valid mapping (target, sources) in the text editor before i can manage it via the graphical perspective.

It should be possible to build a mapping from scratch using the graphical perspective without resorting to code perspective.
![image](https://github.com/user-attachments/assets/223bc7e6-60b8-489b-8308-cd437562a991)

harmen-xb:
@willem-xbreeze Same story as for #83 and #85, we need to first know the target object before actually creating the mapping. I think you can also create a new mapping by right clicking on an entity, then it's know and the object model should be fine.

Let's decide generically in story https://dev.azure.com/x-breeze/CrossModel/_workitems/edit/913 on object creation flow and then implement it consistently for this and related issues.

harmen-xb:
Fixed as part of #115, the mapping is now created using a dialog asking what the target entity is.

PR #115, opened 2025-03-13
Object creation with dialogues

colin-grant-work:
This PR adjusts the current dialog system to request all required information before creating a file.

It is an intermediate stage between the current custom dialog system and one based around reuse of the form widget to allow users to create new objects without triggering errors that make the form unusable.

## To do
 - [x] Make sure UI tests pass.
 > They behave a little differently locally, so I'm creating the PR to get CI running. Will ping when ready for review.

### DIFF
--- a/e2e-tests/src/tests/form/add-edit-delete-relationship.spec.ts
+++ b/e2e-tests/src/tests/form/add-edit-delete-relationship.spec.ts
@@ -46,8 +46,9 @@ test.describe('Add/Edit/Delete relationship from explorer', () => {
       const editor = await app.openCompositeEditor(NEW_RELATIONSHIP_PATH, 'Code Editor');
       expect(await editor.textContentOfLineByLineNumber(1)).toBe('relationship:');
       expect(await editor.textContentOfLineByLineNumber(2)).toMatch('id: NewRelationship');
-      expect(await editor.textContentOfLineByLineNumber(3)).toMatch('parent:');
-      expect(await editor.textContentOfLineByLineNumber(4)).toMatch('child:');
+      expect(await editor.textContentOfLineByLineNumber(3)).toMatch('name: "NewRelationship"');
+      expect(await editor.textContentOfLineByLineNumber(4)).toMatch('parent: Address');
+      expect(await editor.textContentOfLineByLineNumber(5)).toMatch('child: Customer');
       await editor.close();
    });
 
--- a/e2e-tests/src/tests/model/add-delete-model.spec.ts
+++ b/e2e-tests/src/tests/model/add-delete-model.spec.ts
@@ -20,9 +20,9 @@ async function confirmCreationDialog(app: CMApp, entityName: string, modelType: 
 
 test.describe.serial('Add/Edit/Delete model from explorer', () => {
    let app: CMApp;
-   const NEW_MODEL_PATH = 'testFolder/NewModel';
-   const NEW_MODEL_PACKAGE_PATH = 'testFolder/NewModel/package.json';
-   const NEW_MODEL2_PATH = 'testFolder/NewModel2';
+   const NEW_MODEL_PATH = 'testFolder/NewDataModel';
+   const NEW_MODEL_PACKAGE_PATH = 'testFolder/NewDataModel/package.json';
+   const NEW_MODEL2_PATH = 'testFolder/NewDataModel2';
    test.beforeAll(async ({ browser, playwright }) => {
       app = await CMApp.load({ browser, playwright });
    });
@@ -40,11 +40,11 @@ test.describe.serial('Add/Edit/Delete model from explorer', () => {
       if (!tabBarToolbarNewModel) {
          return;
       }
-      const name = 'NewModel';
-      const modeltype = 'logical';
+      const name = 'new-data-model';
+      const modelType = 'logical';
       const version = '0.0.1';
       await tabBarToolbarNewModel.trigger();
-      await confirmCreationDialog(app, 'NewModel', 'logical', '0.0.1');
+      await confirmCreationDialog(app, name, modelType, version);
 
       // Verify that the model was created as expected
       await explorer.activate();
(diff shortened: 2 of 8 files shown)


harmen-xb:
@colin-grant-work Understand the playwright tests will fail because the creation is now going through a dialog.

If it takes a lot of time to make it work (>2 hours), I would choose to disable and park these tests till we refactored to using the form on more forgiving grammar setup (which I discussed with @martin-fleck-at this afternoon).

We can accept manual testing these interactions for the coming weeks, since the first thing for April will be implementing the new creation flow anyway.

colin-grant-work:
@harmen-xb, thanks for the pass :-). It looks like only one test was failing, so it was a quick job to fix them, fortunately.

colin-grant-work:
That said, I'm not sure _how_ some of the tests are passing - e.g. the entity tests are looking for the text `NewEntity`, but they should be seeing `NewLogicalEntity`.

colin-grant-work:
Never mind, we set the text `NewEntity` as part of the test, so that explains that :-).

colin-grant-work:
@harmen-xb, should be, yes. I believe all creation flows are working without validation errors.

harmen-xb:
@colin-grant-work @martin-fleck-at 

I reviewed the changes and this implementation solves issue #81. That's nice, but not the full scope of the story yet.

The key problem we had, is that files were created before the name (and thus id) of the object is known. This was the case in the file explorer and the system diagram.

The implementation up till now solves it for the file explorer, but not the diagram.

For example, when I create an entity in a system diagram the file is immediately created (NewEntity1.entity.cm for example), before I have entered a functional name and pressed enter. So the file name is incorrect from the start. This problem only exists for Entity objects in the system diagram (a.t.m.).
When creating relationships in the system diagram there is already logic which derives the relationship-name based on the parent and child (which should be equal behavior as when using the dialog and not entering a Name).

From both starting points (file explorer and system diagram) the end-user will fill in the name (and not the id) of the object. That name should be translated into an id (so strip whitespaces and remove special chars) and use the id as the base for the filename (so add extension).

**Requested changes:**

- [ ] I notice the name field in the dialog now is very restrictive (it doesn't allow spaces and special characters), can you change this to conform to the grammar?
- [ ] For object types which only require a name, the dialog doesn't specify a label for the input. Can you add it so it's clear the 'Name' is asked?
- [ ] Add logic to 'Create Entity' diagram tool so entity files are created after the name is entered (and derive id + filename based on it)

colin-grant-work:
> Add logic to 'Create Entity' diagram tool so entity files are created after the name is entered (and derive id + filename based on it)

Am I right that this is the crux of the matter? Basically, you're happy with the behavior from the file tree entry point, but would like the same behavior when entering the entity creation flow from a diagram?

> I notice the name field in the dialog now is very restrictive (it doesn't allow spaces and special characters), can you change this to conform to the grammar?

Happy to switch this, but I want to make sure that I understand correctly, as I don't believe I changed the validation logic. Here it is on [main](https://github.com/CrossBreezeNL/crossmodel/blob/efc3b7db39ef5d543d73b777d867eaf6dae76b1a/packages/core/src/browser/new-element-contribution.ts#L291-L302), and here it is [before my first PR](https://github.com/CrossBreezeNL/crossmodel/commit/faaacdbee95958000d6342b7b10c67273b6cbc71#diff-bea2b60672c143a0abe4e1667eca35cfe7de0ef86b37d965726fe8f9380c6bdaL235-L237). But the goal now is to allow users to input names conforming to the grammar for names and then transform them appropriately for the ID field?

colin-grant-work:
> - [x] I notice the name field in the dialog now is very restrictive (it doesn't allow spaces and special characters), can you change this to conform to the grammar?
> - [x] For object types which only require a name, the dialog doesn't specify a label for the input. Can you add it so it's clear the 'Name' is asked?
> - [ ] Add logic to 'Create Entity' diagram tool so entity files are created after the name is entered (and derive id + filename based on it)

@harmen-xb, I've pushed one commit that addresses the second of your requests and ensures that all the dialogues have a label for the 'Name' part and a second that addresses the first, that we treat the input as a name - very free in its form - and derive the ID from it. I consider the second commit a bit riskier, since I'm less sure about what structure we want to impose on the name.

Regarding the third request, to add the same flow to the entity creation from diagrams, I've asked @martin-fleck-at to have a look at it. Things behave differently in the diagram because we currently create the file before we get any user input at all so that the entity exists for the diagram to refer to - otherwise it goes locks up because of an invalid reference. I'm not as familiar with the GLSP side of things, so Martin has a better chance of getting it done in time for you to have a look at it before the demo.

harmen-xb:
@colin-grant-work Thanks for the update, entering the name feels much more user friendly.

I tested creating all object types, and the creation works but there one issue:

- When creating a relationship the parent and child drop-downs only work when creating the relationship in a folder named "relationships". So for new data models I have to manually create a folder named "relationships" and then right click on that folder and create the new relationship in order to get the drop-downs populated with the entities in the local model.

colin-grant-work:
> When creating a relationship the parent and child drop-downs only work when creating the relationship in a folder named "relationships".

Hm... that would be an odd condition. The code just queries the server for referenceable entities in scope. It's possible that the scope computation at the top level (i.e. without any subfolders) goes wrong. I'll take a look at that.

colin-grant-work:
@harmen-xb, it looks like the current scope computation does not recognize that the directory in which a `package.json` resides belongs to the scope of that `package.json`. To me, that seems incorrect, but a naive attempt to change that logic results in some test failures in the completion provider. I can probably fix the symptom by providing a dummy child path for the documents we're planning to create - would that be alright for the current requirements?

colin-grant-work:
Or maybe it doesn't cause issues with the tests - after my first run failed, subsequent runs passed. Could have been a funny state of the build.

harmen-xb (approved):
@colin-grant-work Thanks! Now it works as expected.

PR #116, opened 2025-03-17
Use Manhattan routing for system and mapping diagram

martin-fleck-at:
- Upgrade to nightly GLSP for bug fixes regarding routing
- Ensure sizing of system nodes
- Avoid storing routings for mapping diagram


harmen-xb:
@martin-fleck-at Thanks for the routing change, looks better :).

I noticed it wasn't added for inheritance, so I added it there as well.

While testing I also noticed a serialization problem (which I guess isn't due to changes in this branch), but it is the reason I looked at the grammar and found 1 small thing (the IdentiifedFragment wasn't used). This should have an impact, and the serialization problem remains.

So here are steps to reproduce the issue:
- Open the mapping-example workspace
- Open the ExampleDWH > ProductInheritance system diagram, the diagram should look fine
- Now move the 'Product' entity a bit and let go. All inheritance lines disappear and problems appear in the problems list.

What happened in the file is the following, before the moved the edges part looks like this:
image

And after the move it looks like this:
image

See how the id property of the edges is moved to last position. I can't find where the order of the properties is defined in the serialization, but I guess you do :). Could you have a quick look?

Thanks!

harmen-xb (approved):
Reviewed code and implementation and it works correct now, thanks!

PR #124, opened 2025-03-28
Feature/less restrictive grammar

harmen-xb:
As preparation for using the forms for creating new objects:
- Refactored grammar to be less restrictive (allow empty id's and made all properties optional).
- Added id property to customProperties and refactored implementation and examples accordingly.
- Refactored ErrorView and disabled properties on forms to only activate on parse errors.
- Refactored validations to rely on IdentifiedObject supertype i.s.o. generic AstNode parent.
- Refactored unique id validation to validate semantic root objects globally and all other IdentifiedObject types locally (within collection).
- Resolved lint warning on deprecated WebSocketConnectionProvider usage.

### DIFF
--- a/extensions/crossmodel-lang/src/glsp-server/mapping-diagram/handler/create-edge-operation-handler.ts
+++ b/extensions/crossmodel-lang/src/glsp-server/mapping-diagram/handler/create-edge-operation-handler.ts
@@ -41,7 +41,7 @@ export class MappingEdgeCreationOperationHandler extends JsonCreateEdgeOperation
          const sourceElement = this.modelState.index.findSemanticElement(sourceElementId, isSourceObjectAttribute);
          const targetElement = this.modelState.index.findSemanticElement(targetElementId, isTargetObjectAttribute);
          const sourceEntity = <LogicalEntity | undefined>getOwner(sourceElement);
-         if (!targetElement || !targetElement.id || !sourceElement || !sourceElement.id || !sourceEntity || !sourceEntity.id) {
+         if (!targetElement?.id || !sourceElement?.id || !sourceEntity?.id) {
             return;
          }
          const sourceAttributeReference = combineIds(sourceEntity.id, sourceElement.id);
--- a/extensions/crossmodel-lang/src/glsp-server/mapping-diagram/handler/create-edge-operation-handler.ts
+++ b/extensions/crossmodel-lang/src/glsp-server/mapping-diagram/handler/create-edge-operation-handler.ts
@@ -12,7 +12,7 @@ import {
 } from '@eclipse-glsp/server';
 import { inject, injectable } from 'inversify';
 import { combineIds } from '../../../language-server/cross-model-naming.js';
-import { isSourceObjectAttribute, isTargetObjectAttribute, LogicalEntity } from '../../../language-server/generated/ast.js';
+import { isSourceObjectAttribute, isTargetObjectAttribute, SourceObject } from '../../../language-server/generated/ast.js';
 import { createAttributeMapping, createAttributeMappingSource, getOwner } from '../../../language-server/util/ast-util.js';
 import { CrossModelCommand } from '../../common/cross-model-command.js';
 import { MappingModelState } from '../model/mapping-model-state.js';
@@ -40,11 +40,11 @@ export class MappingEdgeCreationOperationHandler extends JsonCreateEdgeOperation
          const container = this.modelState.mapping.target;
          const sourceElement = this.modelState.index.findSemanticElement(sourceElementId, isSourceObjectAttribute);
          const targetElement = this.modelState.index.findSemanticElement(targetElementId, isTargetObjectAttribute);
-         const sourceEntity = <LogicalEntity | undefined>getOwner(sourceElement);
-         if (!targetElement?.id || !sourceElement?.id || !sourceEntity?.id) {
+         const sourceObject = <SourceObject>getOwner(sourceElement);
+         if (!targetElement?.id || !sourceElement?.id || !sourceObject?.id) {
             return;
          }
-         const sourceAttributeReference = combineIds(sourceEntity.id, sourceElement.id);
+         const sourceAttributeReference = combineIds(sourceObject.id, sourceElement.id);
          const existingMapping = container.mappings.find(mapping => mapping.attribute?.value.ref?.id === targetElement.id);
          if (existingMapping) {
             existingMapping.sources.push(createAttributeMappingSource(existingMapping, sourceAttributeReference));
--- a/extensions/crossmodel-lang/src/glsp-server/mapping-diagram/handler/create-edge-operation-handler.ts
+++ b/extensions/crossmodel-lang/src/glsp-server/mapping-diagram/handler/create-edge-operation-handler.ts
@@ -1,7 +1,7 @@
 /********************************************************************************
  * Copyright (c) 2024 CrossBreeze.
  ********************************************************************************/
-import { TARGET_ATTRIBUTE_MAPPING_EDGE_TYPE, isLeftPortId, isPortId } from '@crossbreeze/protocol';
+import { isLeftPortId, isPortId, TARGET_ATTRIBUTE_MAPPING_EDGE_TYPE } from '@crossbreeze/protocol';
 import {
    Command,
    CreateEdgeOperation,
@@ -12,7 +12,7 @@ import {
 } from '@eclipse-glsp/server';
 import { inject, injectable } from 'inversify';
 import { combineIds } from '../../../language-server/cross-model-naming.js';
-import { isSourceObjectAttribute, isTargetObjectAttribute } from '../../../language-server/generated/ast.js';
+import { isSourceObjectAttribute, isTargetObjectAttribute, LogicalEntity } from '../../../language-server/generated/ast.js';
 import { createAttributeMapping, createAttributeMappingSource, getOwner } from '../../../language-server/util/ast-util.js';
 import { CrossModelCommand } from '../../common/cross-model-command.js';
 import { MappingModelState } from '../model/mapping-model-state.js';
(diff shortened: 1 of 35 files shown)


harmen-xb:
@martin-fleck-at I made some changes as you can see as preparation for the in-memory new object form story.

While refactoring the grammar I made the id property value optional, so the form can be shown although the id & name properties are missing or empty. For this I had to change some code in the language-server & glsp-server. This part I am the least confident about whether these changes are the right way to handle the optional'ness of the properties. See commit https://github.com/CrossBreezeNL/crossmodel/pull/124/commits/aaf5ebc6cc6c47e8a5114bcc947eccd6c9594bf9.

I tested the flow in CrossModel and it all seems to work fine.

Let me know you're comments/feedback.

packages/protocol/src/model-service/protocol.ts
```diff
@@ -172,64 +177,57 @@ export interface AttributeMapping extends CrossModelElement, WithCustomPropertie
...
    readonly $type: typeof SourceObjectAttributeType;
 }
 
 export const SystemDiagramType = 'SystemDiagram';
-export interface SystemDiagram extends CrossModelElement, Identifiable, WithCustomProperties {
+export interface SystemDiagram extends CrossModelElement, IdentifiedObject, WithCustomProperties {
    readonly $container: CrossModelRoot;
    readonly $type: typeof SystemDiagramType;
-   customProperties: Array<CustomProperty>;
-   description?: string;
```
martin-fleck-at, line 199: Is it on purpose that we are removing the description and the name from the diagram?

extensions/crossmodel-lang/src/language-server/grammar/common.langium
```diff
@@ -3,20 +3,20 @@ import 'terminals'
 // Common interfaces and fragments.
 
 interface IdentifiedObject {
-    id: string;
+    id?: string;
 }
 
 fragment IdentifiedObjectFragment returns IdentifiedObject:
-    'id' ':' id=ID;
+    'id' ':' (id=ID)?;
```
martin-fleck-at, line 10: nit: Is there a reason this parenthesis is like this compared to all the other optional properties?

extensions/crossmodel-lang/src/language-server/cross-model-validator.ts
```diff
@@ -87,70 +88,78 @@ export class CrossModelValidator {
...
+      }
       if (isSystemDiagram(node)) {
          this.markDuplicateIds(node.edges, accept);
          this.markDuplicateIds(node.nodes, accept);
       }
       if (isMapping(node)) {
          this.markDuplicateIds(node.sources, accept);
       }
+      if (isWithCustomProperties(node)) {
+         this.markDuplicateIds(node.customProperties, accept);
```
martin-fleck-at, line 161: This does not seem to work for custom properties within the target mappings of a mapping diagram.

extensions/crossmodel-lang/src/language-server/cross-model-validator.ts
```diff
@@ -87,70 +88,78 @@ export class CrossModelValidator {
...
-         accept('error', 'The name of this object cannot be empty', { node: namedObject, property: 'name' });
+         accept('error', 'The name cannot be empty', { node: namedObject, property: 'name' });
+      }
+   }
+
+   checkIdentifiedObject(identifiedObject: IdentifiedObject, accept: ValidationAcceptor): void {
+      if (identifiedObject.id === undefined || identifiedObject.id.length === 0) {
+         accept('error', 'The id cannot be empty', { node: identifiedObject, property: 'id' });
+      } else {
+         // Only perform the following checks when the id is know.
```
martin-fleck-at, line 99: nit:
```suggestion
         // Only perform the following checks when the id is known
```

extensions/crossmodel-lang/src/glsp-server/mapping-diagram/handler/create-edge-operation-handler.ts
```diff
@@ -40,11 +40,12 @@ export class MappingEdgeCreationOperationHandler extends JsonCreateEdgeOperation
          const container = this.modelState.mapping.target;
          const sourceElement = this.modelState.index.findSemanticElement(sourceElementId, isSourceObjectAttribute);
          const targetElement = this.modelState.index.findSemanticElement(targetElementId, isTargetObjectAttribute);
-         if (!targetElement || !sourceElement) {
+         const sourceEntity = <LogicalEntity | undefined>getOwner(sourceElement);
```
martin-fleck-at, line 43: nit: I think the typing here is wrong. The owner of a source object attribute should be the source object, not an entity. So it would be more like: `const sourceEntity = getOwner<SourceObject>(sourceElement);` if we adapt one of the overloaded function calls to allow generic, i.e., `export function getOwner<T extends AstNode>(node?: AstNode): T | undefined;`

extensions/crossmodel-lang/src/glsp-server/mapping-diagram/handler/create-edge-operation-handler.ts
```diff
@@ -40,11 +40,12 @@ export class MappingEdgeCreationOperationHandler extends JsonCreateEdgeOperation
...
+         if (!targetElement || !targetElement.id || !sourceElement || !sourceElement.id || !sourceEntity || !sourceEntity.id) {
```
martin-fleck-at, line 44: nit: Could be shortened to:
```suggestion
         if (!targetElement?.id || !sourceElement?.id || !sourceEntity?.id) {
```

martin-fleck-at (changes requested):
Hi @harmen-xb, thank you very much for your change! As far as I can tell, everything is still working as expected and also the server adaptations are correct. I only have a few minor comments and I noticed that the custom properties ID check does not work in all cases (see inline comment). Great work!

packages/protocol/src/model-service/protocol.ts
harmen-xb: This was already removed from the grammar, so I made it in sync now.

extensions/crossmodel-lang/src/language-server/grammar/common.langium
harmen-xb: I first had the parenthesis around the whole block (including the 'id' part) but then langium doesn't seem to like it. Cause then the frangment can be empty.

extensions/crossmodel-lang/src/glsp-server/mapping-diagram/handler/create-edge-operation-handler.ts
harmen-xb: Got it, fixed in the code. getOwner was already generic.

harmen-xb:
@martin-fleck-at I processed your comments, fixed the custom properties validation issue and also fixed a couple of serialization issues I encountered while testing.

colin-grant-work:
@harmen-xb, with this PR, I'm seeing the following behavior on `main`

1. Create new entity via dialog.
2. In the form editor, delete the name.
3. Observe an error toast
> Request textDocumentClientCapabilities/documentSymbol failed.

It seems that in this code in langium: 

https://github.com/eclipse-langium/langium/blob/c5de9ba5fd21fe60a0071f3b20b9004e3520baf4/packages/langium/src/lsp/document-symbol-provider.ts#L50-L57

This code in CrossModel is being called:

https://github.com/CrossBreezeNL/crossmodel/blob/6bc045597b3c4932bfeefda9524933eced824e64/extensions/crossmodel-lang/src/language-server/cross-model-naming.ts#L120-L122

That code is finding the ID node, with range (0 based) `{start: {line: 1, character: 8}, end: {line: 1, character: 24}`, but the parent node, allegedly representing the whole `LogicalEntity` has a range `{start: {line: 0, character: 0}, end: {line: 1, character: 0}}` and that's breaking this validation

https://github.com/eclipse-theia/theia/blob/b7b06f08a5d28302863d7f4d2809f63ae7abca1f/packages/plugin-ext/src/plugin/types-impl.ts#L2168-L2170

because the range of the ID node isn't contained in the range of the complete node.

harmen-xb:
Hi @colin-grant-work ,

Thanks for the research.

Based on your message it sounds like a bug in langium, or are we using the parent functionality the wrong way?

martin-fleck-at:
@colin-grant-work I can only reproduce this if I click on the error in the Problems view, is that the same for you? It seems weird to me that the parent does not contain the child nodes but not having a "name" property node on an entity that does not have a name seems reasonable and we should probably produce the error on the `namedObject` directly without any property since it does not exist:

https://github.com/CrossBreezeNL/crossmodel/blob/6bc045597b3c4932bfeefda9524933eced824e64/extensions/crossmodel-lang/src/language-server/cross-model-validator.ts#L92

(same for id)

What do you think?

colin-grant-work:
> Based on your message it sounds like a bug in langium, or are we using the parent functionality the wrong way?

I agree with this, but I'm not sure whether the issue is on our side or theirs: it does seem that it shouldn't be possible to have any kind of parent node that doesn't span the full range of its children.

> I can only reproduce this if I click on the error in the Problems view, is that the same for you?

No, I see the problem as soon as the name is deleted, whether that's in the form editor or the text editor; I don't have to interact with the problems in any way.

> not having a "name" property node on an entity that does not have a name seems reasonable.

I agree with this, and I haven't followed the logic on our side very closely to see whether returning the `id` instead of the `name` is really intended, but

> It seems weird to me that the parent does not contain the child

to me, this seems like the fundamental problem. Regardless of whether we want to return the `id` or nothing in this particular case, the `LogicalEntity` CST node's range should include all of its attributes; it shouldn't cut off at the end of the word `entity:`.

martin-fleck-at:
Ah, I can reproduce it now. Seems to be a bug in Langium where dedents are not flushed at the correct end position but rather at the first column of the last line, so if we do not have any new lines at the end of our file, this will result in wrong range calculation. I provide a workaround here https://github.com/CrossBreezeNL/crossmodel/pull/127 and will create an issue in Langium some time soon.

PR #125, opened 2025-03-31
feat: Object creation with in-memory editors

colin-grant-work:
This PR makes a number of changes to support the creation of objects through an untitled form editor rather than through a dedicated dialog.

- Changes on the server side to support client-only documents with non-file URI's
  - Allow the client to send initial text, rather than reading from disk.
  - Manage lifecycle of client-only resources: delete if closed on client.
  - Package discovery for non-file URI's
  - Update to the GLSP relationship creation logic to use the same naming scheme
- Changes on the UI side to open an untitled editor / form for Relationships and LogicalEntities
  - At the moment, only those two use this mechanism for new object creation, as they both rely on form editors. In principle, it should be possible to do the same for diagrams, but it will require some work on the diagram side to provide initial text, etc.
- Some adjustments to save logic to allow the user to save the new object in a file whose name and ID are derived from the user's `name` input without going through a separate 'Save As' dialog.
- Adjustments to the language server's `CrossModelValidator` and the `RelationshipForm` to provide better feedback in error cases that don't cause parse errors.
  - Required fields (Name, Parent, Child) should be highlighted if absent.
  - Attribute fields should be highlighted and have appropriate error text on hover if erroneous (e.g. attribute used more than once)
  - If the name field is currently set to the default that would computed from parent and child and the editor is untitled, then the name should update automatically if the parent or child is changed; otherwise the name should remain as the user input it.

Some of the modifications will carry over to other forms, but my focus has been on the Relationship form, as suggested by the user story.

I do have a couple of questions regarding specific scenarios:
 - What should happen in case the user attempts to create a new relationship in a context in which there are not two referenceable entities? It would be possible to throw an error early and warn the user that the context doesn't support relationship creation, or to let the user open the form and perhaps then create new entities with which to populate it.
 - Relatedly, the user story suggests starting with a completely blank slate. In the current state, defaults are provided for parent, child, and (derived) name. From a UX perspective, I think that's preferable, as it means the user starts out in an error-free state, perhaps even their desired state, and it's very hard to get into an erroneous state from the form itself, other than by deleting the name field.
 - The ID is hidden while the form is untitled, and then ID and eventual filename are computed on save. In the untitled state, `NewRelationship` or `NewLogicalEntity` are used as the title for the editor's tab. That's potentially confusing, but updating the title to reflect the computed state could be quite distracting and would differ from the behavior of all other tabs. If you have any preferences for how the tab should be presented, please let me know.


e2e-tests/src/tests/cm-tabbar-toolbar.spec.ts
```diff
@@ -20,72 +20,6 @@ test.describe('CrossModel TabBar Toolbar', () => {
       await explorer.tabBarToolbar.waitForVisible();
    });
 
-   test('create new entity from tabbar toolbar', async () => {
```
colin-grant-work, line 23: These two tests were redundant, as the same behavior of creation from tabbar is tested separately in the entity and relationship test suites.

martin-fleck-at:
@harmen-xb There is a test case failing in the CI that we cannot reproduce locally. Could you please chech whether you have more success in that?
```
Error:   1) diagram/system/add-edit-delete-attributes.spec.ts:60:8 › Add/Edit/Delete attributes to/from an entity in a diagram › Edit attribute via properties view 

    Retry #1 ───────────────────────────────────────────────────────────────────────────────────────
    Error: expect(received).toMatchObject(expected)

    - Expected  - 2
    + Received  + 2

      Object {
        "datatype": "Boolean",
    -   "description": "New Description",
    -   "identifier": true,
    +   "description": "",
    +   "identifier": false,
        "name": "Renamed Attribute",
      }

      73 |       // Verify that the attribute is changed in the properties view
      74 |       const properties = await attribute.getProperties();
    > 75 |       expect(properties).toMatchObject({
         |                          ^
      76 |          name: RENAMED_ATTRIBUTE_LABEL,
      77 |          datatype: 'Boolean',
      78 |          identifier: true,
        at /home/runner/work/crossmodel/crossmodel/e2e-tests/src/tests/diagram/system/add-edit-delete-attributes.spec.ts:75:26
Warning: e2e-tests/model/add-delete-model.spec.ts took 15.1s
Notice:   1 failed
    diagram/system/add-edit-delete-attributes.spec.ts:60:8 › Add/Edit/Delete attributes to/from an entity in a diagram › Edit attribute via properties view 
```

harmen-xb:
> -edit-delete-attributes.spec.ts:60:8 › Add/

I can reproduce it locally:

image

harmen-xb:
When I rebase this branch on latest main the issue is gone locally, so I pushed the rebase here as well.

willem-xbreeze:
> ~When I rebase this branch on latest main the issue is gone locally, so I pushed the rebase here as well.~
> 
> Apparently it ran succesfully once, after that different scenarios fail. They are all scenarios where the MUI Datagrid is used and I am wondering whether it is related to it being setup so that it's effectuated once you click outside of the datagrid.
> 
> So if the testcase is creating/changing something in a datagrid and immediately checking the code it might nog be in the file yet. As aligned with @martin-fleck-at if this is the problem, the datagrid implementation can be changed that the change is effectuated immediately (cause it caused more issues while modeling anyway).

Related issue:
https://github.com/CrossBreezeNL/crossmodel/issues/135

harmen-xb:
@colin-grant-work @martin-fleck-at Thanks for the implementation, nice to see I can create objects before the file is there.

I do have some findings/questions:
- When I create a new entity and don't save it yet (so it's not persisted) and then create a new relationship I can already select the new entity as a child or parent. This shouldn't happen, relationships can only be created on actually created objects.
- When I create a new relationship the parent and child fields are already selected. Can we change it so nothing is selected initially and when the parent and child fields are selected the name is updated accordingly (unless the users has manually changed the name). If for example the initial content of the file is like this the form works fine:
  ```
  relationship:
      id: NewRelationship
      name: ""
  ```
- I notice the id property is pre-populated and computed on save. Is it possible to update the id and filename property on any name change? So that when I save the state of the window doesn't change (the id and filename were already correct).

colin-grant-work:
@harmen-xb, I had left a couple of questions in the description, and it sounds like you've answered most of them. There is still this one:

>What should happen in case the user attempts to create a new relationship in a context in which there are not two referenceable entities? It would be possible to throw an error early and warn the user that the context doesn't support relationship  creation, or to let the user open the form and perhaps then create new entities with which to populate it (this is the current behavior).

Let me make sure that I understand your answers to these:

> Relatedly, the user story suggests starting with a completely blank slate. In the current state, defaults are provided for parent, child, and (derived) name. From a UX perspective, I think that's preferable, as it means the user starts out in an error-free state, perhaps even their desired state, and it's very hard to get into an erroneous state from the form itself, other than by deleting the name field.

> [!NOTE]
> From your requests, it sounds like you'd prefer the blank slate approach. That raises two additional questions:
> - Would you like errors to be hidden from the form view when expected (i.e. don't mark parent as missing until it has previously been present?
> - Would you like saving to be prevented while errors are present, or is it fine to save in an error state of some kind, and then proceed to fix them on the saved.

> The ID is hidden while the form is untitled, and then ID and eventual filename are computed on save. In the untitled state, `NewRelationship` or `NewLogicalEntity` are used as the title for the editor's tab. That's potentially confusing, but updating the title to reflect the computed state could be quite distracting and would differ from the behavior of all other tabs. If you have any preferences for how the tab should be presented, please let me know.

> [!NOTE]
> Here, it sounds like you'd like the tab to update to reflect the new name, whenever we compute it? And also, then, whenever the user modifies it?

> The autosave functionality should be disabled for these in-memory editors. It should not allow to autosave untill the user explicitly saved the file once. Currently when enabling autosave I end up with 2 files, the autosaved file and the file which is created on explicit save.

Hm... it should already have been disabled automatically for untitled URI's, so I'm surprised it's kicking in, but with the Theia uplift, we've also brought in some [recent changes](https://github.com/eclipse-theia/theia/blob/30188778390f92744ae8d8c5f3d53054abce4124/packages/core/src/common/resource.ts#L65-L66) to the saveable system that should make it trivial to mark our untitled editors as not autosaveable.

harmen-xb:
> @harmen-xb, I had left a couple of questions in the description, and it sounds like you've answered most of them. There is still this one:
> 
> > What should happen in case the user attempts to create a new relationship in a context in which there are not two referenceable entities? It would be possible to throw an error early and warn the user that the context doesn't support relationship  creation, or to let the user open the form and perhaps then create new entities with which to populate it (this is the current behavior).

With the blank slate approach I think this isn't relevant. If one creates a relationship in a model without entities, the child and parent dropdowns are empty.

> 
> Let me make sure that I understand your answers to these:
> 
> > Relatedly, the user story suggests starting with a completely blank slate. In the current state, defaults are provided for parent, child, and (derived) name. From a UX perspective, I think that's preferable, as it means the user starts out in an error-free state, perhaps even their desired state, and it's very hard to get into an erroneous state from the form itself, other than by deleting the name field.
> 
> Note
> 
> From your requests, it sounds like you'd prefer the blank slate approach. That raises two additional questions:
> 
> * Would you like errors to be hidden from the form view when expected (i.e. don't mark parent as missing until it has previously been present)?

I don't mind the errors for now. I guess it's extra work to hide them before the first save?

> * Would you like saving to be prevented while errors are present, or is it fine to save in an error state of some kind, and then proceed to fix them on the saved.

Saving in error state is fine.

> 
> > The ID is hidden while the form is untitled, and then ID and eventual filename are computed on save. In the untitled state, `NewRelationship` or `NewLogicalEntity` are used as the title for the editor's tab. That's potentially confusing, but updating the title to reflect the computed state could be quite distracting and would differ from the behavior of all other tabs. If you have any preferences for how the tab should be presented, please let me know.
> 
> Note
> 
> Here, it sounds like you'd like the tab to update to reflect the new name, whenever we compute it? And also, then, whenever the user modifies it?

Yes, if easily possible. Modifying the file to reflect the actual id with which it will store with is more important I think. The file shouldn't change when saving in my opinion (which is happening now for the id property).

> 
> > The autosave functionality should be disabled for these in-memory editors. It should not allow to autosave untill the user explicitly saved the file once. Currently when enabling autosave I end up with 2 files, the autosaved file and the file which is created on explicit save.
> 
> Hm... it should already have been disabled automatically for untitled URI's, so I'm surprised it's kicking in, but with the Theia uplift, we've also brought in some [recent changes](https://github.com/eclipse-theia/theia/blob/30188778390f92744ae8d8c5f3d53054abce4124/packages/core/src/common/resource.ts#L65-L66) to the saveable system that should make it trivial to mark our untitled editors as not autosaveable.

colin-grant-work:
@harmen-xb, I've pushed a couple of commits. One changes the ID along with the name when the name, parent, or child are changed. The second fixes autosave behavior. Theia recently consolidated that functionality into a service so that individual editors no longer have to implement autosaving, so I've removed the code from `model-editor` that was handling that. While working on that, I found two odd behaviors in Theia:
 - Untitled editors couldn't be closed without saving if the autosave preference was not `off`. I've put in a fix for that.
 - When closing untitled editors, you're prompted to save. If you save, a new editor is opened, because that's the default behavior for 'Save As' cases. I've added some code that handles that in a bit of a roundabout way with a `Proxy`. If that seems odd, I can also:
   - Just leave the behavior as in Theia, where the editor will be opened.
   - Override the relevant method more directly, copy over most of the code from Theia, but omit the call to open the editor where appropriate.

harmen-xb (approved):
@colin-grant-work Changes look good, I made some small changes in the last commit after playing with creating entities and relationships a bit.

PR #127, opened 2025-04-08
Fix bug in Langium where dedents are flushed at the wrong position

martin-fleck-at:


colin-grant-work (approved):
With these changes, I'm unable to reproduce the problem, but other parsing and validation behavior looks correct. 👍

PR #128, opened 2025-04-09
Feature/publish packages

harmen-xb:
- Updated lerna to latest version and adapted build scripts
- Renamed all packages to be prefixed with @crossbreezenl scope
- Added automatic versioning, tagging and release creation in workflows.
- Refactored feature (non-PR) commits to only run on ubuntu
- Disabled Docker image build on main


willem-xbreeze (approved):

harmen-xb:
Hi @colin-grant-work @martin-fleck-at ,

This is a message to inform you both that we integrated our cicd changes into main, including a rename of all packages (where the scope needed to be changed from @crossbreeze to @crossbreezenl so we can publish to the github package repository.

These changes also include an automatied mechanism for sematic versioning using the conventional commits specification (https://www.conventionalcommits.org/en/v1.0.0/). So in order for the versioning to work correctly, could you include the right keyswords in the commits when fixing bugs, added functionalities and making breaking changes?

Thanks!

PR #130, opened 2025-04-10
Allow user to specify name before creating a new entity

martin-fleck-at:


harmen-xb:
@martin-fleck-at Thanks for the change, it works good.

One question, in the object creation flow story we talked about have field be shown in the diagram canvas at the position where was clicked to enter the name. Something like:
image


Is this easily possible? Cause I think from end-user this is more user friendly.

martin-fleck-at:
@harmen-xb You are absolutely correct that this would have been a bit nicer and that was what I was aiming for first. Unfortunately, it requires more infrastructure (because we need the validation cycle) and also is a bit harder to deal with since we are always in continuous mode (i.e., the user can just continue to clicking on the diagram and create new entities) and we would need to prevent that or assume that it is canceling the input. Since the dialog gave us all of that for free, I went for the "cheaper" option but I can look into it deeper if you want. Now as I am writing this, maybe it already help to simply position the dialog a bit better and make it look more like a single text input through CSS. We still need to have it modal (I believe) but it might lead to nicer UX. Will look quickly into it.

packages/glsp-client/src/browser/system-diagram/node-creation-tool/system-node-creation-tool.ts
```diff
@@ -37,8 +50,40 @@ export class SystemNodeCreationToolMouseListener extends NodeCreationToolMouseLi
...
+            action.args.name = name;
+            this.tool.dispatchActions([action]);
+         });
+         return MessageAction.create('', { severity: 'NONE' });
       }
       throw new Error('Invalid node creation type');
    }
+
+   protected async queryEntityName(): Promise<string | undefined> {
+      const referenceableEntities = await this.tool.modelService.findReferenceableElements({
```
harmen-xb, line 68: When testing the new dialog I noticed when I enter 'Order' in a model where the Order entity does not exist, but it does exist in another model in the workspace the id will be Order1 i.s.o. Order. I'm not sure this line is involved in it, but it looks like the scope for checking names is not set to the local model, but to the workspace.

martin-fleck-at: I pushed an update that should fix that

harmen-xb: That UI change a great improvement! Works nicely now.

harmen-xb (approved):
Changed look good!

PR #139, opened 2025-04-14
feat:Upgrade Theia & GLSP

colin-grant-work:
Updates Theia dependencies to 1.60.2 and GLSP versions to:
 - `[latest](https://www.npmjs.com/package/@eclipse-glsp/theia-integration/v/2.4.0-theia.1.60.0)` for `@eclipse-glsp/theia-integration` for Theia 1.60 compatibility.
 - Current next for remaining `@eclipse-glsp` dependencies, except those for which `@eclipse-glsp/theia-integration/` specifies specific versions.

> [!NOTE]
> We could instead resolve those dependencies to the current `next` rather than the lower `2.4.0` versions specified by the Theia integration package. Doing so doesn't cause any UI test breakage or regressions that I noticed.

A few subsidiary adjustments were required:
 - Any `p-*` references to Phosphor classes changed to `lm-*` after Theia's migration to LuminoJS.
 -  `@eclipse-glsp/glsp-playwright` changed export style for `asLocator` function.
 - Theia 1.60 changed argument ordering for `EditorManager::revealSelection`.


harmen-xb (approved):
@colin-grant-work Thanks for the upgrade!

Changes all look good to me. While testing I also didn't notice anything strange, so I will merge it.

harmen-xb:
> @harmen-xb it looks like you've already merged this, but in the future, how would like the prefixes to work? I usually orient myself toward the `Squash and Merge` PR merge option, where the PR title becomes the final commit message, so I've applied the `feat:` prefix there; if you prefer that it be part of the branch name or 'raw' commit message(s), or all three or some subset, just let me know.

The chore branch is fine, I just wasn't aware of this category. It's added in the workflow now.

I also prefer the squach and merge. I just noticed the versioning went wrong for some reason, while it worked on the PR I did earlier. The only difference is 'feat:bla' versus 'feat: Bla' space after feat:. So I adjusted the object creation PR name to see whether this solves it.

PR #140, opened 2025-05-08
Update system and mapping diagram lines

martin-fleck-at:
- Use libavoid on client-side to improve edge routing experience
- Use ELK on server side to layout system diagram (ALT+L)
- Make ports visible in Mapping diagram


martin-fleck-at:
@harmen-xb I think I fixed everything we discussed, please feel free to have a look.

harmen-xb:
@martin-fleck-at That's a big improvement! Looks a lot better. Nice to see the soruce and target attribute hovering also works now.

I do notice when I create a new mapping using 'Derive mapping' the target node doesn't seem to be position correctly on the canvas (or the canvas is rendering weirdly):

image

I notice when I drag the canvas down it stays correct.

martin-fleck-at:
@harmen-xb I think this is because we allowed negative y coordinates if there is no source object in the layout. I could reproduce this easily by pressing the 'Reset Viewport' button in the toolbar. I pushed a commit that should fix that.

harmen-xb (approved):
Changes look good to me.

PR #144, opened 2025-05-26
fix: Allow same global ids if types are different

### LINKED ISSUE #134: Cannot have a diagram with the same id as an entity
If i have an entity with id Customer and trying to create a diagram with id customer i get an error.
It should be possible to have multiple objects of different types with the same id.

Image

martin-fleck-at:
Fixes #134


harmen-xb (approved):
Looks good, tested and works as expected.

PR #146, opened 2025-05-26
fix: Ensure no error is thrown if there are no sources in a mapping

### LINKED ISSUE #143: When creating mapping using  'New Element > Mapping' the target attributes are not in the file
When creating a new mapping using the right-click menu > New Element > Mapping the mapping file is created but the target attributes are missing (so the file is invalid).

When doing the same but right clicking on an entity and derive mapping the mapping is created correctly (but not in the right folder yet).

martin-fleck-at:
Fixes #143

### DIFF
--- a/extensions/crossmodel-lang/src/language-server/cross-model-validator.ts
+++ b/extensions/crossmodel-lang/src/language-server/cross-model-validator.ts
@@ -179,7 +179,7 @@ export class CrossModelValidator {
       this.markDuplicateIds(node.customProperties, accept);
    }
 
-   protected markDuplicateIds(nodes: IdentifiableAstNode[], accept: ValidationAcceptor): void {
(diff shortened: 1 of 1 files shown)


harmen-xb:
@martin-fleck-at I still see the issue.

When I have an entity with one or more attributes and I right click on the entity and choose 'Derive mapping' I get a mpping where the target attributes are already listed:

```
mapping:
    id: NewMapping2
    target:
        entity: Address
        mappings:
            - attribute: Customer_Id
            - attribute: Street
            - attribute: CountryCode
```

But if I right click on the model and choose 'New element > Mapping' and choose the same target entity the target attributes are not in the file (and the mapping diagram is not rendered correctly):

```
mapping:
    id: NewMapping
    target:
        entity: Address
```

martin-fleck-at:
Sure, I'll have a look at this and merge it with the PR in https://github.com/CrossBreezeNL/crossmodel/pull/147

PR #149, opened 2025-06-10
fix: Avoid validating package.json as it is not a language file

martin-fleck-at:
Fixes https://github.com/crossmodel/crossmodel-core/issues/118


harmen-xb (approved):
LGTM

Issue #102, opened 2025-02-14
Broken link in the readme


Bas-xb:
The link to the "Prerequisites Windows" in the readme under "Getting started" is broken

https://github.com/CrossBreezeNL/crossmodel/blob/main/docs/PrerequisitesWindows..md
(2 dots instead of 1)

PR #151, opened 2025-06-10
chore: Updated Mui, Langium & other dependencies

harmen-xb:
- Updated Mui to 7.1.1 and refactored toolbar in GridComponent.
- Updated Langium to 3.5.0.
- Updated some other minor dependencies
- Fixed error in link in README (#102)

### DIFF
--- a/extensions/crossmodel-lang/src/language-server/cross-model-workspace-manager.ts
+++ b/extensions/crossmodel-lang/src/language-server/cross-model-workspace-manager.ts
@@ -1,9 +1,9 @@
 /********************************************************************************
  * Copyright (c) 2023 CrossBreeze.
  ********************************************************************************/
-import { AstNode, DefaultWorkspaceManager, Deferred, FileSelector, FileSystemNode, LangiumDocument } from 'langium';
+import { AstNode, DefaultWorkspaceManager, Deferred, FileSelector, FileSystemNode, LangiumDocument, UriUtils } from 'langium';
 import { CancellationToken, Emitter, Event, WorkspaceFolder } from 'vscode-languageserver';
-import { URI, Utils } from 'vscode-uri';
+import { URI } from 'vscode-uri';
 import { CrossModelSharedServices } from './cross-model-module.js';
 
 /**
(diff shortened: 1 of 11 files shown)


extensions/crossmodel-lang/src/language-server/cross-model-workspace-manager.ts
```diff
@@ -60,7 +60,7 @@ export class CrossModelWorkspaceManager extends DefaultWorkspaceManager {
          return name !== 'out';
       } else if (entry.isFile) {
          const extname = Utils.extname(entry.uri);
-         return fileExtensions.includes(extname);
+         return selector.fileExtensions.includes(extname);
```
martin-fleck-at, line 63:
```suggestion
         return selector.fileExtensions.includes(extname) || selector.fileNames.includes(Utils.basename(entry.uri));
```

In the newer version Langium also supports selection based on filenames and they added the second check in their code base as well or maybe even call `super.includeEntry` in case it is a file since we do not have any customization there.

martin-fleck-at (changes requested):
Hi @harmen-xb,

Thank you very much for the upgrade! The change looks good to me for the most part. We are just missing the filename handling in Langium (see inline comment) and I noticed one other, minor issue: When adding a new attribute (or any other new row) in a Grid, the `singleSelect` fields have a margin of 16px applied to the top:

image.

harmen-xb:
@martin-fleck-at I looked at the margin issue, and I can't find why this happens.
Nothing changes in how we use the Mui that should influence the margin. Since in the end we want to replace Mui with another component anyway I rather not fix this right now. Takes to much time to find the cause a.t.m.

martin-fleck-at:
I just had a very quick look and I think we could fix it like this in CSS:
```css
/* While editing, nested forms seem to have a wrong margin applied. */
.MuiDataGrid-cell--editing .MuiFormControl-root {
    margin-top: 6px;
}
```

It may also not be perfect though.

martin-fleck-at (approved):

Issue #97, opened 2025-02-06
Changing order of attributes doesn't always go well


Bas-xb:
![Image](https://github.com/user-attachments/assets/7c07196f-ad8f-43c7-92d6-bebec60796dd)

Steps to reproduce:
1. Create a new entity
2. Add 2 attributes with different datatypes
3. Change the order of either of the attributes. 

Now the name of one of the attributes is copied over the other instead of changing the order.
However, in the code editor both attributes are still there but the datatype is wrong.

![Image](https://github.com/user-attachments/assets/bc925006-924a-4250-bd83-55cdc839519e)

![Image](https://github.com/user-attachments/assets/1f9f1419-5ddf-4e9b-8f44-8f2fc7552897)

harmen-xb:
@Bas-xb I can't reproduce this in the #125 PR result. Can you retest this once this is merged?

harmen-xb:
@Bas-xb Changes have been pushed to main now which seem to have fixed the issue you had.

Can you verify?

Bas-xb:
@harmen-xb 
I can still reproduce this issue with the steps I posted above

PR #153, opened 2025-06-14
feat: Add datamodel to crossmodel language

harmen-xb:
- Added datamodel concept to langium grammar definition
- Added example datamodel files in mapping example workspace.


martin-fleck-at:
@harmen-xb I pushed a commit on this PR to switch from package.json to datamodel.cm and introduce a form to manage the data model and it's dependencies.

martin-fleck-at (approved):
I approve the PR so that it is approved from my side for your changes but I think it would be good if you could also look at the changes I made.

harmen-xb:
@martin-fleck-at Changes look good, definitiely a lot better then editing a package.json file :).

I do notice something strange (to be honest I also saw this in another branch but it's pretty anoying):
- When you open two objects of the same type, so a datamodel file from one model and then form the second. I you jump back to the first one the form shown is from the second. Only after closing all forms and opening on the one I want to edit works. So it looks like some state is shared across the tabs when working in the same form.
The same happens for entities and relationships. I think this is introduced recently, cause I am pretty sure this wans't an issue 2 months ago.

martin-fleck-at:
@harmen-xb You are right! I pushed a commit that should fix that issue!

harmen-xb:
@martin-fleck-at That's better indeed, so it was a MUI issue.

Seems isue #97 is also solved with this PR.

PR #171, opened 2025-09-15
feat: Add existing relationship to system diagram

martin-fleck-at:
- Support dragging relationship files onto the system diagram
- Add tool for adding existing relationships to the system diagram


harmen-xb:
@martin-fleck-at I reviewed the workings in CrossModel and it works good, also with adding entity nodes when they are not there yet, or using entities when they are.

I did notice one issue:
- When using the 'Show Entity' tool it also lists relationship. This should only be entities.

harmen-xb (approved):
LGTM!

PR #173, opened 2025-09-18
chore: Frontend migrate mui to primereact

### LINKED ISSUE #137: Need to click "Add attribute" when creating multiple attributes
Open the form on an entity.
Click Add Attribute, a new attribute is added.
Click Add Attribute again, nothing happens (only attribute grid exits edit mode)
Click Add attribute again, a new attribute is added.

Every click on add attribute should add a new attribute.

harmen-xb:
Refactoring of MUI front-end components to PrimeReact.

This refactoring also introduces filters on data grid components (like entity attributes)

Fixes #137 #135 #97


packages/react-model-ui/src/views/common/RelationshipAttributesDataGrid.tsx
```diff
@@ -2,189 +2,484 @@
...
+            child: attribute.child || '_',
+            id: newId,
+            idx: -1
+         };
+
+         dispatch({
+            type: 'relationship:attribute:add',
+            attribute: attributeData
+         });
+         setEditingRows({ [newId]: true });
```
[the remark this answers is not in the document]
harmen-xb: @hardiksinh Please remove the default _ values so the values are undefined by default.
Only commit the relationship attribute when a child and/or attribute has been set. Otherwise discard the row on stop editing.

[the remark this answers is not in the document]
hardiksinh: @harmen-xb I understand, but I set default _ because PrimeReact autocomplete fields by default setting _ when no value selected. Another problem is, we directly add row on click of Add Attribute and then open in Edit mode. There is no state like Draft Mode. Hence, the problem is even if nothing is changed and exit edit mode, the row remains with _ values for both child and parent. I am trying to bind event and remove raws where this is default value, but unable to fix yet. Similar issue with other AutoComplete fields which don't requires save. For example, SourceObject and DataModel.

packages/react-model-ui/src/views/common/DataModelCustomPropertiesDataGrid.tsx
```diff
@@ -2,51 +2,113 @@
...
+}
 
 export function DataModelCustomPropertiesDataGrid(): React.ReactElement {
    const dataModel = useDataModel();
    const dispatch = useModelDispatch();
    const readonly = useReadonly();
+   const [editingRows, setEditingRows] = React.useState<Record<string, boolean>>({});
+   const [validationErrors, setValidationErrors] = React.useState<Record<string, string>>({});
+
+   const validateField = React.useCallback((rowData: CustomPropertyRow): Record<string, string> => {
```
martin-fleck-at, line 22: I believe this has been discussed before but I just want to note it down here: Ideally validation comes from the server and already has a proper message so we do not have to deal with it on the client side.

packages/react-model-ui/src/views/common/Icons.tsx
```diff
@@ -3,14 +3,16 @@
...
-import React = require('react');
+import * as React from 'react';
 
-export const KeyIcon: React.FC<SvgIconProps> = props => (
+interface IconProps {
+   className?: string;
+   style?: React.CSSProperties;
+}
+
+export const KeyIcon: React.FC<IconProps> = props => (
```
martin-fleck-at, line 13: This seems to be unused? Either we should remove the file or we render the icon as part of the column header in the entity attributes.

packages/react-model-ui/style/primereact-overrides.css
```diff
@@ -0,0 +1,460 @@
...
+
+/* PrimeReact component styles */
+
+/* Theme-specific styles */
+.light-theme {
+   --surface-ground: #ffffff;
+   --surface-section: #f8f9fa;
+   --surface-card: #ffffff;
+   --surface-border: #dfe7ef;
+   /* --text-color: #000000; */
```
martin-fleck-at, line 24: Any particular reasons they are commented out?

packages/react-model-ui/src/views/form/Header.tsx
```diff
@@ -22,36 +22,36 @@ export function Header({ name, id, iconClass }: HeaderProps): React.ReactElement
    const untitled = useUntitled();
    const diagnostics = useDiagnostics();
 
+   const startContent = (
```
martin-fleck-at, line 25: nit: Would be nice to also move some of the style definitions here to CSS.

packages/react-model-ui/src/views/common/PrimeDataGrid.tsx
```diff
@@ -0,0 +1,498 @@
...
+      };
+
+      document.addEventListener('mousedown', handleClickOutside);
+      return () => {
+         document.removeEventListener('mousedown', handleClickOutside);
+      };
+   }, [activeRowKey]);
+
+   // eslint-disable-next-line react/prop-types
+   const allActionsTemplate = (rowData: T, props: ColumnBodyOptions): React.JSX.Element => {
```
martin-fleck-at, line 302: When I adapt the text in the text editor so we get an invalid model, e.g., changing `id` property to `ixd`, the actions are still visible and clickable instead of being disabled.

martin-fleck-at (changes requested):
I reviewed the code and did a few tests and most of it looks good to me. Great work @hardiksinh! I do have some questions or nitpicks (which may be ignored) that I left in inline comments. Other than that, I just have a few styling and usability concerns but obviously @harmen-xb has the last word on those ;-)

- Overlays are not properly styled. This is true for the filter overlay but also for selection/options, e.g., Data Type:
overlay-styling

- Selected checkbox looks a bit odd:
checkbox-styling

- When looking at the mapping source object conditions, numbers are not properly rendered. References and regular text works fine:
number-error

- Error message stays even if the value changes, at least in the datamodel dependency grid:
dependency-error

- There are a few places where we use `setTimeout` which is always a bit of a risk as it may depend on the browser or host machine whether the timeout is sufficient or not. This can also lead to unstable tests. So far I did not encounter any issues but it is something to bear in mind.

- In general, I think the spacing in the form is a bit odd. It seems quite crowded but that is just my personal opinion.

I think all those things can actually be done in a follow-up if we want this PR merged to avoid any merge conflicts. Just let me know and I'll change my feedback to 'Approve'.

extensions/crossmodel-lang/src/main.ts
```diff
@@ -19,18 +18,24 @@ import { createConnection, ProposedFeatures } from 'vscode-languageserver/node.j
  * - Create a RPC-based model server that exposes an API to access the Langium AST/semantic model on a dedicated port, e.g., for form-access
  */
 
-// Create a connection to the client
-const connection = createConnection(ProposedFeatures.all);
+async function main(): Promise<void> {
```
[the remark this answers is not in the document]
hardiksinh: When I run build for browser, it was giving me below error because mixing of CommonJS (require) and ESM (import) in project. For which I needed to change the code. 

`Error [ERR_REQUIRE_ESM]: require() of ES Module /var/www/html/crossmodel/packages/server/lib/index.js from /var/www/html/crossmodel/extensions/crossmodel-lang/out/main.cjs not supported. Instead change the require of index.js in /var/www/html/crossmodel/extensions/crossmodel-lang/out/main.cjs to a dynamic import() which is available in all CommonJS modules. at Object.<anonymous> (/var/www/html/crossmodel/extensions/crossmodel-lang/out/main.cjs:9988:21) { code: 'ERR_REQUIRE_ESM' }`

packages/react-model-ui/src/ThemedViewer.tsx
```diff
@@ -1,172 +1,59 @@
...
+const darkTheme: ThemeStyles = {
+   '--primary-color': '#0e639c',
+   '--primary-color-text': '#ffffff',
+   '--surface-ground': '#1e1e1e',
+   '--text-color': '#d4d4d4',
+   '--surface-card': '#252526',
+   '--surface-border': '#404040',
+   '--surface-hover': '#2a2d2e',
+   '--focus-ring': '0 0 0 0.2rem #0e639c40'
+} as ThemeStyles;
```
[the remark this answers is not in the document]
hardiksinh: Sure, let me remove unnecessary type assertions for both lightTheme and darkTheme objects.

packages/react-model-ui/src/views/common/DataModelCustomPropertiesDataGrid.tsx
hardiksinh: Okay, I will check.

hardiksinh:
> I reviewed the code and did a few tests and most of it looks good to me. Great work @hardiksinh! I do have some questions or nitpicks (which may be ignored) that I left in inline comments. Other than that, I just have a few styling and usability concerns but obviously @harmen-xb has the last word on those ;-)
> 
>     * Overlays are not properly styled. This is true for the filter overlay but also for selection/options, e.g., Data Type:
> 
Can you please let me know which browser/OS version you are using? I could not replicate in Firefox/Chrome/Edge in Ubuntu.

>     * Selected checkbox looks a bit odd:
> 
I see, the problem with Firefox browser. I will fix it.

>     * When looking at the mapping source object conditions, numbers are not properly rendered. References and regular text works fine:
> 
I do not understand problem I think. Can you please give me an example of what it shows and what is expected?

>     * Error message stays even if the value changes, at least in the datamodel dependency grid:
> 
Yes, it is known issue, I need to fix.

>     * There are a few places where we use `setTimeout` which is always a bit of a risk as it may depend on the browser or host machine whether the timeout is sufficient or not. This can also lead to unstable tests. So far I did not encounter any issues but it is something to bear in mind.
> 
Yes, there are few on click events. I will check and try fixing to get it updated without setTimeout. 
>     * In general, I think the spacing in the form is a bit odd. It seems quite crowded but that is just my personal opinion.
> 
It is actually what we wanted to make forms as compacts as possible as @harmen-xb requested. 
> 
> I think all those things can actually be done in a follow-up if we want this PR merged to avoid any merge conflicts. Just let me know and I'll change my feedback to 'Approve'.

I will push few updates as you mentioned in other comments, and you may finally review before approve.

hardiksinh:
> I also see the overlay issue when I switch to dark theme color (ignore the mouse pointer position, I clicked on the filter icon):

I fixed it now by adding dark-theme class to body tag and styling in overrides css, as overlays load outside the PrimeReact component. I try adding both light and dark theme stylesheet of PrimeReact in override and modifying ThemedViewer.tsx to load stylesheet dynamically, so that we can remove lot of overrides and keep only the necessary modification to actual layout. But I could not correctly set it up to apply the dark theme. It always set light theme. Probably, PrimeReact loading of style tags dynamically and here trying different way is not working. But for now this overrides will work. And we may need to fix things which may look odd.
If you can spend a little time, please check PrimeReact theme switching implementation and how to replace in our code.

packages/react-model-ui/src/views/common/PrimeDataGrid.tsx
hardiksinh: @harmen-xb Okay, will use the same.

martin-fleck-at:
@hardiksinh Changes work for me! Thank you for the quick turnaround!

